PR #7, opened 2018-04-19
fix(webpack): prevent hmr chunks from being inserted into the template

robin-drexler:
Same issue as in https://github.com/xing/hops/pull/466


dmbch:
Thanks! 🚀

PR #16, opened 2018-04-27
Refactor Express and Webpack Mixins

dmbch:


packages/webpack/mixin.core.js
```diff
@@ -204,17 +211,22 @@ class WebpackMixin extends ExpressMixin {
...
-    Object.assign(this.options, argv);
+    this.options = { ...this.options, ...argv };
   }
 }
 
 WebpackMixin.strategies = {
   configureWebpack: pipe,
   inspectBuild: sequence,
+  build: override,
+  clean: override,
```
robin-drexler: Couldn't `cleanup` also be of type `sequence`? This would allow mixins to clean up their own stuff without having to know and worry about others.

dmbch: I initially thought both `clean` and `build` ought to be called in `parallel` to cater for advanced requirements. I am convinced that they ought to be work similarly: users would (assumably) want to control both in a consistent way - they probably would only have to extend or alter the `clean` implementation if they also modified `build`.

Coming up with a parallelizable `build` implementation, however, is non-trivial, as we would have to find a way to modify stats handling (currently used for logging). I would not want to handle logging directly inside the `build` function - and I cannot think of an acceptable way to pick the stats object from an array that we would receive if we switched to `parallel` strategy.

So, for now, I propose to merge and iterate if and when the need arises. Currently, users/implementers can already subclass `WebpackMixin` to provide their own bespoke methods. @robin-drexler, does that work for you?

robin-drexler: Absolutely. Thanks for the elaborated explanation! :)

PR #19, opened 2018-05-02
[RFC] React: allow to hook into template with head and body functions

robin-drexler:
This PR is intended as a discussion starter.

Projects such as [styled-components](https://www.styled-components.com/docs/advanced#server-side-rendering) or [react-loadable](https://github.com/jamiebuilds/react-loadable#loadablecapture) (both links lead directly to their respective SSR docs) analyze the rendered app in order to gather information. styled-components finds the used styles on the page. react-loadable determines the components that have been loaded in the request.

However, currently besides overriding `render`, there seems not to be a way to alter the template data after `ReactDom.renderToString` has been called.

## Idea
Allow to add markup to head/body with hooks (`prependHead`, `appendHead`, `prependBody`, `appendBody`)  without having to override the entire template or render function.
Those hooks will be called *after* `ReactDom.renderToString` so they'd be able to access any data created by react-loadable, styled-components or similar packages.

### DIFF
--- a/packages/react/lib/template.js
+++ b/packages/react/lib/template.js
@@ -8,6 +8,7 @@ module.exports = data =>
   `<!DOCTYPE html>
 <html ${data.helmet.htmlAttributes.toString()}>
   <head>
+    ${data.prependHead().join('')}
     ${data.helmet.title.toString()}
     ${data.helmet.base.toString()}
     ${data.helmet.meta.toString()}
@@ -15,11 +16,15 @@ module.exports = data =>
     ${data.assetsByType.css.map(cssLink).join('')}
     ${data.helmet.style.toString()}
     ${data.helmet.script.toString()}
+    ${data.appendHead().join('')}
+    
   </head>
   <body ${data.helmet.bodyAttributes.toString()}>
+    ${data.prependBody().join('')}
     <div id="${data.mountpoint}">${data.markup}</div>
     ${data.helmet.noscript.toString()}
     ${data.globals.map(variable).join('')}
     ${data.assetsByType.js.map(jsLink).join('')}
+    ${data.appendBody().join('')}
   </body>
 </html>`.replace(/(^\s*[\r\n]| (?=>))/gm, '');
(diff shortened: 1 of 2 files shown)


dmbch:
Thanks for your proposal! I generally really like the idea, I only have a couple of questions:

* do we really have to add this many hooks to achieve what we want? why can we not just expose `renderToString` as a mixing hook and make it `override`-able?
* would it alternatively/additionally make sense to introduce a new generic `enhanceData` hook (`sync.pipe`) to allow for such data manipulations after rendering?
* why not (optionally) pass the additional template data as plain strings to the template function instead of functions returning arrays?

robin-drexler:
> do we really have to add this many hooks to achieve what we want? why can we not just expose `renderToString` as a mixin hook and make it `override`-able?

Wouldn't this only allow one mixin to add data? Admittedly I also would not know how one could add template data in such a hook, despite it getting passed `data` and returning a tupel of the rendered string and the mutated `data`

> would it alternatively/additionally make sense to introduce a new generic `enhanceData` hook (`sync.pipe`) to allow for such data manipulations after rendering?

[That was the first idea I tried actually](https://github.com/untool/untool/compare/fetch-data-after-render) but I refrained from opening a PR. 
Although this approach would help in the react-loadable case, where one needs to add js files to `assetsByChunkName`, it would still not allow to add markup to the head as needed for styled-components (inline css). 

> why not (optionally) pass the additional template data as plain strings to the template function instead of functions returning arrays?

I'm not entirely sure how that'd look like. Such function would need to be called after `renderToString`, which would kind of turn it into `enhanceData`, right?

robin-drexler:
Thinking about it. My proposed approach might not be the most convenient to add to `data` either.

How about a `enhanceData`, which is called after `renderToString` , but also add `startHead` and `endHead` (same for body) attributes to `data` which can be added to in the hook.

This would allow to add chunks but also freestyle html. 

Maybe that was your proposal all along and I just did not get it at first.

dmbch:
Yes, exactly - that was what I meant to suggest in questions two and three. Sorry for being unclear...

PR #20, opened 2018-05-03
Add enhanceData React Mixin Hook

dmbch:
Introduce a new `@untool/react` mixin hook to allow to augment template data after fetching. This is especially useful for server side rendering, in which case it will be called after the call to `ReactDOM.Server`'s `renderToString` method.

Resolves #19


dmbch:
Re: 404 middleware - I thought it was somewhat fitting, as it essentially makes the `Miss` component work as expected...

Anyway, I see your point. :)

PR #25, opened 2018-05-08
Fix node_externals mechanism

dmbch:


packages/webpack/lib/utils/helpers.js
```diff
@@ -33,6 +33,25 @@ exports.checkESNext = (target, defaults) => {
   return check;
 };
 
+exports.determineExternals = (target, defaults) => {
+  const resolve = createResolver(exports.getResolveConfig(target, defaults));
+  const checkESNext = exports.checkESNext(target, defaults);
+  const isNotExternal = (context, request) => {
+    try {
+      return request.startsWith('.') || checkESNext(resolve(context, request));
```
ZauberNerd: What about `startsWith('/')` or `startsWith('~')`?

ZauberNerd: Also: You are now creating a resolver in multiple places. Here in [line 37](https://github.com/untool/untool/pull/25/commits/9efc5cc4cd9426a3736daf6dc447662a4e449687#diff-725bed128020e90199ba7a494eefda14R37) and also inside `checkESNext()` in [line 22](https://github.com/untool/untool/blob/9efc5cc4cd9426a3736daf6dc447662a4e449687/packages/webpack/lib/utils/helpers.js#L22).
Maybe we could change the `checkESNext` function to only accept absolute paths and create the resolver only in one place?

packages/webpack/lib/utils/helpers.js
```diff
@@ -33,6 +33,25 @@ exports.checkESNext = (target, defaults) => {
   return check;
 };
 
+exports.determineExternals = (target, defaults) => {
+  const resolve = createResolver(exports.getResolveConfig(target, defaults));
+  const checkESNext = exports.checkESNext(target, defaults);
+  const isNotExternal = (context, request) => {
```
ZauberNerd: Please rename this function. How about `shouldBundle` or something similar?

packages/webpack/lib/utils/helpers.js
ZauberNerd: Another case where this won't work: Adding custom module dirs to the webpack config, via `configureWebpack`, because this resolver will always use the default config instead of the customized webpack config.

dmbch: As I am not quite sure how to handle resolver config (using @untool's config mechanism), I would like to postpone `moduleDirs` handling (including `~`) - and, to be honest, I do not quite get what is special about namespaced modules in this context. Would you mind elaborating on that..?

Your other points, I completely agree with. I will add `path.isAbsolute(request)` and `request.includes('!')` checks to that badly named inner function. Oh, and it turns out, we can get rid of the resolver in `checkESNext()` because it now always receives absolute paths.

PR #27, opened 2018-05-14
WIP: Tweak extensibility

dmbch:


packages/express/mixin.core.js
```diff
@@ -8,37 +8,20 @@ const { Mixin } = require('@untool/core');
 const uri = require('./lib/uri');
 
 class ExpressMixin extends Mixin {
-  create(method) {
-    const create = require(method === 'static'
-      ? './lib/static'
-      : './lib/serve');
+  createServer(mode) {
+    const create = require(mode === 'static' ? './lib/static' : './lib/serve');
```
ZauberNerd: How about writing this like: `const create = mode === 'static' ? require('./lib/static') : require('./lib/serve');`?

packages/express/mixin.core.js
```diff
@@ -8,37 +8,20 @@ const { Mixin } = require('@untool/core');
...
     const { options, config, initializeServer, finalizeServer } = this;
-    return create(method, options, config, initializeServer, finalizeServer);
+    return create(mode, options, config, initializeServer, finalizeServer);
   }
-  run(method) {
+  runServer(mode) {
     const run = require('./lib/run');
     const { config, inspectServer, logInfo, logError } = this;
-    const app = this.create(method);
+    const app = this.createServer(mode);
```
ZauberNerd: Why not pass a fully configured `express app` as arguments to the `runServer()` method?

packages/yargs/index.js
```diff
@@ -5,6 +5,8 @@ const createYargs = require('yargs');
 
 const { bootstrap } = require('@untool/core');
 
+exports.Mixin = require('./mixin.core');
```
ZauberNerd: What's the reason for exporting the mixin?
Also what whould happen in a case where I have multiple mixins (core, browser, server)?

packages/express/index.js
```diff
@@ -0,0 +1,17 @@
+module.exports = {
+  get Mixin() {
```
ZauberNerd: Same here, why export the mixin?

packages/express/index.js
```diff
@@ -0,0 +1,17 @@
...
+    return require('./mixin.core');
+  },
+  get uri() {
+    return require('./lib/uri');
+  },
+  get createServer() {
+    return require('@untool/core').bootstrap().createServer;
```
ZauberNerd: Does this work without having to call `handleArguments({})`? In an earlier version of untool we had to do this in order to set up logging correctly (I think it crashed - or just didn't log anything?)

packages/react/index.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  get Mixin() {
```
ZauberNerd: Same here, why export the mixin?
And what about the browser/server mixins?

packages/webpack/index.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+  get Mixin() {
+    return require('./mixin.core');
```
ZauberNerd: Why are you exporting the mixin?

packages/express/index.js
dmbch: True, it does not log anything unless you also call `handleArguments()`. We will probably also have to expose that - or would you consider it appropriate to call it here ourselves?

packages/express/mixin.core.js
dmbch: Certainly possible, but in my opinion, that would somewhat deteriorate our API: one would have to call two methods instead of just one..:

```javascript
const { createServer, runServer } = this;
runServer(createServer(mode));
```

...versus...

```javascript
this.runServer(mode);
```

I personally prefer the latter, but I will make sure you can also use the former: all it takes is a `typeof` check.

packages/express/index.js
dmbch: I think we ought not: `handleArguments` is not just being used to set the log level.

Looking at what [untool/webpack](https://github.com/untool/untool/blob/f21a94533edd8fbd741c16440ddbc7d00ba69408/packages/webpack/mixin.core.js#L220)) is doing, I am not entirely sure if makes a lot of sense to expose these hooks here at all. Will think about it.

packages/express/mixin.core.js
dmbch: Ah, all right. :)

PR #39, opened 2018-06-08
Assorted Tweaks

dmbch:


packages/core/lib/config.js
```diff
@@ -10,40 +10,44 @@ const isPlainObject = require('is-plain-object');
...
-      config.mainFields.splice(1, 0, 'mixin:runtime');
+    switch (target) {
+      case 'core':
+        return resolveCoreMixin(context, mixin);
+      case 'server':
+        return resolveServerMixin(context, mixin);
+      case 'browser':
+        return resolveBrowserMixin(context, mixin);
+      default:
+        throw new Error();
```
ZauberNerd: I'd prefer to either `return null` here in the `default` or throw a more meaningful error.

PR #41, opened 2018-06-12
Rename `namespace` to `name`

### LINKED ISSUE #40: Suggestion: Rename "namespace" to "appName"
Because `namespace` is confusing, especially with our internal tools, where we also use the term `namespace`, but with a different meaning.
/cc @jhiode @matthias-reis

dmbch:
Closes #40


packages/core/lib/config.js
```diff
@@ -113,9 +110,7 @@ const loadPreset = (context, preset) => {
   } catch (error) {
     if (!isResolveError(error)) throw error;
     try {
-      return loadConfig(
-        dirname(resolvePreset(context, join(preset, 'package.json')))
-      );
+      return loadConfig(dirname(resolve(context, `${preset}/package.json`)));
```
ZauberNerd: Please use `path.join()` instead of hardcoding a `/`

dmbch: Since this is essentially a module path, I think it is more sensible to go for a `/` in this case.

PR #44, opened 2018-06-15
refactor(react): simplify, remove components

dmbch:


packages/react/mixin.server.js
```diff
@@ -56,28 +53,25 @@ class ReactPlugin extends Mixin {
...
-              routerContext.headers && res.set(routerContext.headers);
-              res.end();
-            } else {
-              res.status(routerContext.status || 200);
-              routerContext.headers && res.set(routerContext.headers);
-              res.type('html');
-              res.send(document);
-            }
-          });
+        const { context = {} } = this.options;
```
robin-drexler: I kinda liked the name `routerContext` as it gave a clear differentiation to React's `context`

packages/react/mixin.server.js
```diff
@@ -56,28 +53,25 @@ class ReactPlugin extends Mixin {
...
+        if (context.miss) {
```
robin-drexler: Does this `staticContext` based API (`miss` being true/false, `url` causing a redirect etc.) have any tests and or documentation?

I just wonder how anyone could know about them. The components provided a higher level API that potentially made it easier to use.

packages/react/mixin.server.js
dmbch: Well, since we are talking about a local variable here (and ‘this.options’ is all about router options), I was thinking the naming was clear/self-documenting enough. It is called ‘context’ in the context of ‘ReactRouter’, anyway :)

packages/react/mixin.server.js
robin-drexler: Absolutely. :)

PR #48, opened 2018-06-20
refactor(react): globals as object hash

ZauberNerd:
This commit changes the type of the `enhanceData` `globals` hook.
The `globals` were previously an array of `{ key: 'x', value: 'y' }`
are now an object hash that can be accessed through `enhanceData` which
will be rendered in the template as list of global variables in a single
script tag.

Also: The globals data is now always considered to be JSON
(no special treatment for dates, functions, regular expressions).

The following input:
```javascript
enhanceData(data) {
    return {
        ...data,
        globals: {
            ...data.globals,
            ...({
                str: 'string',
                num: 0,
                obj: {foo: 'foo'},
                arr: [1, 2, 3],
                bool: true,
                nil: null,
                undef: undefined,
                date: new Date("Thu, 28 Apr 2016 22:02:17 GMT"),
                fn: function echo(arg) { return arg; },
                re: /([^\s]+)/g,
            })
        },
    };
}
```

Will generate this markup:
```html
<script>var str="string",num=0,obj={"foo":"foo"},arr=[1,2,3],bool=true,nil=null,undef=undefined,date="2016-04-28T22:02:17.000Z",fn=undefined,re={};</script>
```

BREAKING CHANGE: `enhanceData.globals` is now an object hash
Before:
```javascript
enhanceData(data) {
    return {
        ...data,
        globals: [
            ...data.globals,
            { name: 'my_globals', value: 'test' },
        ],
    };
}
```

After:
```javascript
enhanceData(data) {
    return {
        ...data,
        globals: {
            ...data.globals,
            my_global: 'test',
        },
    };
}
```


packages/react/lib/template.js
```diff
@@ -1,9 +1,19 @@
 const esc = require('serialize-javascript');
 
-const variable = ({ name, value }) => `<script>${name}=${esc(value)}</script>`;
 const cssLink = (css) => `<link rel="stylesheet" href="/${css}" />`;
 const jsLink = (js) => `<script src="/${js}"></script>`;
 
+const serializeGlobals = (globals) => {
+  var entries = Object.entries(globals);
```
robin-drexler: `var` could be `const` or `let`. Afaik no need to resort to es5 here.

ZauberNerd: Oops, fixed.

ZauberNerd:
@dmbch I didn't think that we use this feature and therefore I disabled it, because according to their readme it also slows down the serialization quite a bit.
But I'm fine with re-enabling it, if we actually need it.

dmbch:
@zaubernerd then let us please just switch it back on. Their README also says it will only be a significant slowdown if used to serialize a lot of data, so it should not concern us.

And you will have to rebase anyway... ;)

ZauberNerd:
@dmbch reworded - PTAL

PR #49, opened 2018-06-20
RFC: add debug output for config / mixins

ZauberNerd:
This commit adds debug output (via the
[debug package](https://www.npmjs.com/package/debug)).

In order to see the debug output the app needs to be started with the
`DEBUG` environment variable set to `untool:*`.
`$ DEBUG=untool:* un start` will print all debug statements starting
with namespace `untool:`.

This commit introduces two debug namespaces:
- `untool:core`
- `untool:config`

Sample output:
screenshot

### DIFF
--- a/packages/core/lib/config.js
+++ b/packages/core/lib/config.js
@@ -1,3 +1,4 @@
+const debug = require('debug')('untool:config');
 const { basename, dirname, join } = require('path');
 
 const {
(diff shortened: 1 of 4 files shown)


robin-drexler:
🚫🐞

dmbch:
As discussed IRL: I really like this, we should really have debug logging all over ‘untool’.

Do we want to add it in other places within this PR or incrementally?

ZauberNerd:
@dmbch I started adding debug calls to other parts of untool as part of this PR, but then it quickly became a mess and I didn't know which ones are actually important right now, so I removed them again.
I think these two are good enough for now and we can add others incrementally when the need arises.

PR #50, opened 2018-06-20
Simplify express

### LINKED ISSUE #47: Make middleware usage independent from load order
## Problem:

Right now all middlewares are registered in the same order as the mixins are loaded.
This prevents users from overriding the order without losing the benefit of presets.
It is also not possible for a user to remove a middleware.

## Solution (proposal):

Define a `registerMiddleware(s)` hook (strategy: sequence) that collects all middlewares from all mixins first and in another step read the config for the set of middlewares that is desired. This would also allow a user to ignore/remove a middleware later on.
This requires that all mixins name their middlewares.

### Example:
Mixin:
```
registerMiddlewares() {
  return {
    'my-middleware': (app) => (req, res, next) => {
      ...
    },
  }
}
```
Config:
```
{
  middlewares: [
    'my-middleware',
  ],
}
```

P.S.: Yes, middlewares is a word 😄

dmbch:
With this PR, I am re-introducing an `optimizeServer` hook and change `finalizeServer` hook strategy to `override`. This should simplify fine-grained middleware order control.

Additionally, I am removing the default `rewrite` middleware, getting rid of the previous 'catchAll' behaviour. `@untool/express` should have no opinion on URL rewriting.

Closes #47


packages/express/README.md
```diff
@@ -38,14 +38,6 @@ This is equivalent to manually setting `$NODE_ENV` before calling the actual com
 $ NODE_ENV=production un serve
 ```
 
-##### `-s` / `--static`
```
robin-drexler: This one is a bit tricky. It makes sense to remove it, since untool does not do anything with it anymore, but what about other mixins wanting to implement custom rewriting for static mode?

They still could use the `static` flag, right?
Thinking about it: There is currently no way for mixins to add command line arguments to untool "core" commands (build, serve) is it?

dmbch: True, but at least you could add them globally, like `@untool/yargs` used to do in `handleArguments`.

ZauberNerd: @dmbch I don't understand your comment - can you explain / add a link to the code?

dmbch: Sure: https://github.com/untool/untool/blob/0bd5c3a47e7f913e90f94acfbe76ebdc6687e9b1/packages/yargs/mixin.core.js#L11

ZauberNerd: @dmbch ok, so it's global arguments - I don't think that that is a solution we should encourage.
So we would have to extend the mixin class and overwrite the `registerCommands` method.. :/

dmbch: Either that, introduce a new command or rely on config to decide wether to rewrite or not.

dmbch:
Will propose another approach later.

PR #51, opened 2018-06-21
Improve `@untool/express` configurability

### LINKED ISSUE #47: Make middleware usage independent from load order

dmbch:
With this PR, I propose introducing a new hook, `configureServer`, that allows us to control middleware order in a very fine-grained manner. `initializeServer` and `finalizeServer` and the previously used URL rewriting middleware are gone for good ([PTAL](https://github.com/untool/untool/pull/51/files/fd18a8238e0d0e3a1a091fdf2d078c09b1a6129d), @jhiode).

To allow for the addition of arguments to existing commands, I introduced another new hook, `configureCommand` ([PTAL](https://github.com/untool/untool/pull/51/files/f2438d81180c39b51eb86b99ed2057b34fdfff61..10d52c7c93d21fcd8d6903dd346a15a516e04024), @robin-drexler).

Next, to keep our APIs consistent, I aligned the name of the `configureWebpack` hook with the other `configureX` hooks: it is now called `configureBuild`. [PTAL](https://github.com/untool/untool/pull/51/files/93371b3fff3a1242ad3588f191b402866f2cc61e..4cfa0018895d4741653930ec3eeecb075c2710fd).

And finally, I added some additional logging in places where it makes sense for me personally during development. [PTAL](https://github.com/untool/untool/pull/51/files/9759f3eee48318ca3334e5d6816b0ba98943e9f4..45debcacd12cd84f8e25681ef0983774df7e5c90).

Closes #47


jhiode:
@dmbch thanks for this, I will take a look today.

jhiode:
I haven't looked at everything yet, but I think the part that should close #47 isn't sufficient.
I tried to use an artificial setup with the new API and failed at least for 2 cases.

1. access log (initial)
2. user auth (session - but needs the cookie parser)
3. cookie parser (parse)
4. json parser (parse)
5. assets (files - but should happen before user auth)
6. app shells (files)
7. mock api (routes)
8. ssr (routes)
9. error handling (final)

A user authentication middleware needs a cookie parser, so parsers should come before session.
And an asset middleware should come before the session middleware as well as we don't need the user auth status here.

My suggestion is to have this stages:

1. initial
2. parse
3. session (only one that actually needs pre- and post- stages IMO)
4. response
5. final

The example would look like this then:

1. access log (initial)
2. cookie parser (parse)
3. json parser (parse)
4. assets (pre-session)
5. user auth (session)
6. mock api (post-session)
7. app shells (response)
8. ssr (response)
9. error handling (final)

Having multiple response handling middlewares could still be problematic as both of these could wan't to support a catch-all like feature. So just sticking an app-shell and a SSR middleware together would not work per se. But not sure whether we want to support that special case at all.

dmbch:
@jhiode: interesting, as the phases I was proposing are based on what loopback are doing, but with `auth` removed and `files` and `routes` swapped.

I dislike the `response` phase name, though. How about going for these phases..:

- initial
- parse
- assets (or stick to files)
- routes
- final

...and I would want to keep the `pre/post` variants around for semantic reasons: do your phase specific preparation and cleanup there.

Our custom auth middleware could go to `preroutes` (or `postassets`, depending on our semantic preference).

jhiode:
Ok, so this is what LoopBack is doing:
phases
As long as the parse phase is for parsing the request body it is ok (maybe even better) to do that after authentication.
But I think static files should be handled es early as possible (as they normally don't depend on the auth state or the parsed body). But app shells should still work differently.

So I think we should switch the `files` and `parse` phases:
- initial
- files
- parse
- routes
- final

The example would look like this then:
1. access log (initial)
2. cookie parser (initial)
3. assets (files)
4. user auth (pre-parse)
5. json parser (parse)
6. mock api (routes)
7. app shells (routes)
8. ssr (routes)
9. error handling (final)

packages/express/lib/serve.js
```diff
@@ -1,29 +1,31 @@
...
+        ...result,
+        ...additions,
+        middlewareOrder: result.middlewareOrder.concat(Object.keys(additions)),
+      };
+    },
+    { middlewareOrder: [] }
+  );
+  const app = configureServer(express(), middlewares, mode);
+  const { middlewareOrder } = middlewares;
+  debug(middlewares);
```
jhiode: It would be great if we could print the middlewares in the order they will be used.

packages/express/README.md
```diff
@@ -48,27 +48,30 @@ $ un serve -s # OR un serve --static
...
 
-This is a mixin hook defined by `@untool/express` that allows you to register Express middlewares and generally do whatever you like with `app`, the [`Application`](https://expressjs.com/en/api.html#app) (or [`Router`](https://expressjs.com/en/api.html#router)) instance it is using under the hood, right after its creation, i.e. before URL rewriting and such.
+This is a mixin hook defined by `@untool/express` that allows you to register Express middlewares and generally do whatever you like with `app`, the [`Application`](https://expressjs.com/en/api.html#app) instance it is using under the hood.
 
-The second argument it is being called with is `target` and it can be one of the following: `develop`, `serve`, or `static`. If `target` is `static`, `app` will be a [`Router`](https://expressjs.com/en/api.html#router) instance, otherwise an [`Application`](https://expressjs.com/en/api.html#app).
+The second argument it is being called with is `middleware`. It is a plain object containing middleware `Array`s sorted into phases: `initial`, `session`, `parse`, `files`, `routes`, and `final`. Additionally, each of these comes with `pre` and `post` variants.
```
jhiode: "The second argument it is being called with is `middleware`."

packages/express/mixin.core.js
```diff
@@ -36,42 +56,40 @@ class ExpressMixin extends Mixin {
       }, {})
     );
   }
+  configureServer(app, middlewares, mode) {
+    if (mode !== 'static') {
+      const helmet = require('helmet');
+      middlewares.initial.push(helmet());
+      middlewares.files.push(this.createStaticMiddleware());
```
jhiode: `if (mode !== 'static')` => `createStaticMiddleware` we might want to rethink about the naming here  😄

packages/express/lib/serve.js
dmbch: `middlewareOrder` is being already being printed, as it is a property of `middlewares` - and, in my opinion, it is quite useful to at least have the number of middlewares defined for each phase.

PR #53, opened 2018-06-25
Improve untool/react

dmbch:
Clean up and streamline internal APIs:

- get rid of `enhanceData` altogether: its naming was stupid and so was the idea of running it client-side
- introduce server-side-only `getTemplateData` hook instead: using 'context' name here to distinguish template rendering data from generic, universal application data
- refactor a bit to clarify the code and make it more idiomatic


robin-drexler:
I'd really love to avoid shadowing the term `context`, especially in the react package. 
How about something `... afterRender`

dmbch:
How about `getTemplateData`? Of course, this thing is being called after `ReactDOM.renderToString`, but before rendering the SSR template...

robin-drexler:
`getTemplateData` sounds really good to me :)

PR #61, opened 2018-06-28
Environmentalize

dmbch:
Introduce a new feature to support runtime configuration using environment variables. Defining arbitrary preset options using this format `[ENV_VAR]` will activate the feature. 

The previous build-time only environment specific mechanism is removed and replaced by this new one: I think it is just too confusing to support and use two apparently similar approaches that work very differently. The old mechanism is easy to implement in specific presets, though.

### Preset

Define whatever preset option using an environment variable placeholder...

```json
{
  "foo": "[FOO]"
}
```

### CLI Call

...and starting your application in an environment in which the corresponding variable is defined...

```bash
FOO=bar un start
```

### Result/Config

...will yield this runtime configuration:

```json
{
  "foo": "bar"
}
```


packages/webpack/lib/utils/loader.js
```diff
@@ -1,23 +1,26 @@
...
+    target === 'server' ? 'expand(".$1")' : '".$1"'
+  );
+
+const getConfigAndMixins = (...args) =>
+  `{ config: ${getConfig(...args)}, mixins: ${getMixins(...args)} }`;
 
-const getModule = (target, config) =>
-  (target === 'server'
+const getModule = (target, config) => {
+  const configAndMixins = getConfigAndMixins(target, config);
```
ZauberNerd: We talked about allowing to configure the loader without mixins - how about adding another option to the loader, for example: `noMixins: true` or something similar, that then uses `getConfig(target, config)` instead of `getConfigAndMixins()`?
That way the resulting bundle won't include all the code required for the mixins.

packages/webpack/lib/shims/browser.js
```diff
@@ -2,14 +2,17 @@
...
 
 (function render() {
-  var entryPoint = require('@untool/entrypoint');
+  let entryPoint = require('@untool/entrypoint');
   if (typeof entryPoint.default === 'function') {
     entryPoint = entryPoint.default;
   }
-  entryPoint(getConfig())();
+
+  const data = getConfigAndMixins();
```
jhiode: What about `const { config, mixins } = getConfigAndMixins();`? 😄

packages/core/lib/env.js
```diff
@@ -0,0 +1,26 @@
+const isPlainObject = require('is-plain-object');
+
+exports.environmentalize = (_config) => {
+  const _env = {};
+  const env = global._env || process.env;
```
jhiode: I would pass this as an argument as the caller knows better what actually should be used.

packages/express/README.md
```diff
@@ -93,15 +93,15 @@ With this method you can render HTML pages for all configured [`locations`](http
...
-| `host`      | `string`           | `process.env.HOST || '0.0.0.0'` |
-| `port`      | `number`           | `process.env.PORT || 8080`      |
-| `locations` | `[string]`         | `[]`                            |
-| `basePath`  | `string`           | `''`                            |
-| `assetPath` | `string`           | `'<basePath>'`                  |
-| `buildDir`  | `string`           | `'<rootDir>/dist'`              |
+| Property    | Type               | Default            |
+| ----------- | ------------------ | ------------------ |
+| `https`     | `boolean`/`Object` | `false`            |
+| `host`      | `string`           | `[HOST]`           |
```
ZauberNerd: Do you see any way we could keep default values with this setup? Or is that something we want to have support for at all?

packages/core/README.md
```diff
@@ -23,20 +23,21 @@ It allows you to set up mixins and pull in presets. Mixins provide extra functio
...
 
 ```json
 {
-  "port": null,
-  "env": {
-    "production": {
-      "port": 12345
-    }
-  }
+  "port": "[PORT]"
```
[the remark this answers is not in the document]
dmbch: That is certainly true in some specific use cases. Using `process.env.PORT` at compile time implies that environments at compile and run time are identical - and that may or may not be true, depending on circumstances.

And this is an actual example, taken from our `@untool/express` preset, so I would like to keep it as-is.

packages/express/preset.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
   https: false,
-  host: process.env.HOST,
-  port: process.env.PORT,
+  host: '[HOST]',
+  port: '[PORT]',
```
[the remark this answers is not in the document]
dmbch: I think we do want I (see above)... :)

packages/core/lib/env.js
dmbch: I disagree - we only have a single runtime mixin container - and we would have the very same code in there, i.e. at the call site. Additionally, I think it is quite neat to have as much of the code we need to implement this feature at one place instead of having parts of it all over the place, i.e. in two more files.

packages/webpack/lib/utils/loader.js
dmbch: As we discussed IRL: I will make sure the config loader does not explode if the target is non-existing. Thus, if we consider the loader with a non-existing or even `undefined` target, that will work as you describe.

packages/express/README.md
dmbch: I do not really think we need (configurable) default values. At one point, I had a working implementation of this feature (`<foo|defaultFoo>`, `[FOO|defaultFoo]`), but in the end, I was not convinced this is an essential feature.

dmbch:
PTAnotherL, @jhiode @ZauberNerd

packages/express/README.md
dmbch: I have now added support for [dotenv](https://www.npmjs.com/package/dotenv), so I would argue we have a rather nice default value mechanism. Do you agree?

PR #65, opened 2018-07-04
fix: default host to '0.0.0.0' in express before listening on it

ZauberNerd:


dmbch:
I decided to remove the default host definition - according to the [docs](https://nodejs.org/api/net.html#net_server_listen_port_host_backlog_callback) it is obsolete, here:

> if host is omitted, the server will accept connections on the unspecified IPv6 address (::) when IPv6 is available, or the unspecified IPv4 address (0.0.0.0) otherwise.

PR #69, opened 2018-07-05
Autodetect presets

dmbch:
If no presets are explicitly configured, find installed presets amongst direct dependencies.

### DIFF
--- a/packages/core/lib/config.js
+++ b/packages/core/lib/config.js
@@ -90,12 +90,17 @@ const loadConfig = (context, config) => {
     : searchSync(context);
 };
 
-const loadSettings = (context) => {
+const loadSettings = (context, { dependencies }) => {
   const result = loadConfig(context);
-  return {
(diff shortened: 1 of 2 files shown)


packages/core/lib/config.js
```diff
@@ -90,12 +90,17 @@ const loadConfig = (context, config) => {
...
-const loadSettings = (context) => {
+const loadSettings = (context, { dependencies }) => {
   const result = loadConfig(context);
-  return {
-    presets: ['@untool/defaults'],
+  const settings = {
     ...(result ? result.config : {}),
   };
+  if (!settings.presets) {
+    settings.presets = Object.keys(dependencies).filter((key) =>
```
jhiode: I would prefer to pass in the dependencies as an array in the first place.

dmbch: Sorry, was a bit too quick: https://github.com/untool/untool/releases/tag/v0.12.0

PR #70, opened 2018-07-05
Refactor internal webpack mixin api

dmbch:


dmbch:
Not that important now...

PR #73, opened 2018-07-06
feat(webpack): add support for `module-browser` field

dmbch:


ZauberNerd:
I don't think it is neccessary right now - because no one implements it: https://github.com/search?l=JSON&q=%22module-browser%22&type=Code
Also the linked issue seems like just an idea that hasn't been implemented anywhere: https://github.com/webpack/webpack/issues/4674

PR #79, opened 2018-07-13
Add native Browserslist support

### LINKED ISSUE #77: Remove "browsers" from config and use "browserslist" directly
[Browserslist is used by many different tools](https://github.com/browserslist/browserslist#browserslist-) and therefore we should remove the field `"browsers"` from our config and instead use the `browserslist` API directly because that allows our consumers to define their browsers list in [any of the supported](https://github.com/browserslist/browserslist#queries) ways and ensures that other tools may also use the specified browsers list.

dmbch:
In addition to getting rid of the `browsers` config option and adding direct `browserslist` support in the way described in https://github.com/untool/untool/issues/77#issuecomment-404630361, I decided to also get rid of the `node` option. Instead, we are now always building for the `current` node version.

If targeting a specific node version (e.g. when building for [AWS Lambda](https://docs.aws.amazon.com/lambda/latest/dg/current-supported-versions.html)), I would argue it is best practice to also use that version during development. This can, of course, be simplified by using [nvm](https://github.com/creationix/nvm) and enforced by specifying said version in `package.json`:

```json
{
  "engines": {
    "node": "8.10"
  }
}
```

Closes #77


packages/webpack/lib/configs/build.js
```diff
@@ -33,7 +34,7 @@ module.exports = function getConfig(config, configureBuild) {
           {
             modules: false,
             useBuiltIns: true,
-            targets: { browsers: config.browsers },
+            targets: { browsers: loadConfig({ path: config.rootDir }) },
```
ZauberNerd: How about keeping backwards compatibility and write it like: `targets: { browsers: config.browsers || loadConfig(...) }`?

dmbch: Seeing that `browserslist` itself goes out of its way to prevent [conflicting configurations](https://github.com/browserslist/browserslist/blob/master/node.js#L250) from lurking, I am somewhat opposed to this.

Imagine you have your own browserslist config in your project and install some preset containing a browsers field: that preset would override your custom config...

dmbch:
Another argument against this is that it will prevent us from providing a preset that specifies supported browsers.

This might prove especially problematic if we really decide to provide DLL builds of shared components: it that case we really want to be able to define the support matrix in a consistent manner.

ZauberNerd:
@dmbch I think most applications will define the `browsers` query themselves and therefore a preset that pre-defines browsers will always be overwritten.
Let's gather some ideas on how we would use DLL builds and talk about this in the office.

dmbch:
Discussed this with @robin-drexler, @jhiode and @systemboogie - outcome was to close this, at least for now. We want to retain the option to deliver browser support matrices as presets.

Issue #82, opened 2018-07-16
Make getConfig overrideable


ZauberNerd:
Idea: Make [`getConfig`](https://github.com/untool/untool/blob/master/packages/webpack/mixin.core.js#L64) an `override`able mixin hook, so that we can override it in `hops-postcss` in order to pass `cssLoaderConfig` to the `loaderConfigs` and `allLoaderConfigs` arguments of the `configureBuild` hook to allow other mixins that implement `configureBuild` easier access to the `cssLoaderConfig`.

Let's discuss if this is a good idea (tm). Off the top of my head this will make `hops-postcss` order dependent..

PR #84, opened 2018-07-17
Expose getBuildConfig hook

robin-drexler:
This will allow to retrieve the build config for later use, in our case to build a jest-preset.


packages/webpack/README.md
```diff
@@ -108,6 +108,10 @@ If you want to intialize a build of your application, you can do so using this u
 
 Using this utility mixin method, you can delete your `buildDir` and all of its contents. It returns a `Promise`.
 
+### `getBuildConfig()` ([override](https://github.com/untool/mixinable/blob/master/README.md#defineoverride))
```
ZauberNerd: This method needs a `target` argument.

packages/webpack/mixin.core.js
```diff
@@ -3,8 +3,8 @@ const { join } = require('path');
 
 const debug = require('debug')('untool:webpack:stats');
 const {
-  sync: { pipe, sequence },
-  async: { override },
+  sync: { pipe, sequence, override },
```
dmbch: I would prefer renaming `override` to `overrideSync`, similar to https://github.com/untool/untool/blob/master/packages/express/mixin.core.js#L2-L3.

dmbch:
@ZauberNerd this would close #84, would it not?

packages/webpack/mixin.core.js
```diff
@@ -3,7 +3,7 @@ const { join } = require('path');
 
 const debug = require('debug')('untool:webpack:stats');
 const {
-  sync: { pipe, sequence },
+  sync: { pipe, sequence, override: overrideSync },
```
dmbch: Sorry if I was being unclear - I meant having both `overrideSync` and `overrideAsync`, as in..:

```javascript
const {
  sync: { sequence, pipe, override: overrideSync },
  async: { override: overrideAsync },
} = require('mixinable');
```

ZauberNerd:
@dmbch yes, it would close #84 and #82 :P

But the question still stands whether it is a good idea to override this method in the `hops-postcss` package.
But I guess we can postpone that discussion until you're back in the office ;)

PR #85, opened 2018-07-17
Remove static argument in develop mode

### LINKED ISSUE #83: Does it make sense that webpack's "develop" command has a "static" option?
I'm currently reviewing / writing readme documentation and I wondered whether the `--static` option for the `development` command makes sense at all as it seems that `--static` is only used when executing `build`?
https://github.com/untool/untool/blob/master/packages/webpack/mixin.core.js#L192

dmbch:
Closes #83.


packages/webpack/mixin.core.js
```diff
@@ -189,14 +188,7 @@ class WebpackMixin extends Mixin {
...
           describe: `Serve ${name} in watch mode`,
-          builder: {
-            static: {
-              alias: 's',
-              default: false,
-              describe: 'Statically build locations',
-              type: 'boolean',
-            },
-          },
+          builder: {},
```
ZauberNerd: I think you can get rid of this line completely.

dmbch: I would rather not: keeping it around will make the `develop` command easier to extend: if other mixins were to add arguments to it, one would otherwise always have to check if `builder` already exists.

dmbch:
@robin-drexler, are you ok with this? I seem to recall you were working with `static` mode.

robin-drexler:
Shouldn't we remove this from the docs, too?

https://github.com/untool/untool/blob/2d599d332f0056bbb0e9405280bb994b3a0df0a1/packages/express/README.md#-s----static

dmbch:
@robin-drexler we definitely should - good catch. That is another package, though 😉

PR #87, opened 2018-07-21
Protect server.js and assets.json

### LINKED ISSUE #86: server.js should not be served
Currently, server.js is served by the static express middleware, because is written into the `dist/` folder.

This should not happen (by default) as this is probably unexpected behavior.

dmbch:
These files should not be accessible publicly.

Closes #86


dmbch:
@robin-drexler understood - I just updated this PR to make sure things work as you proposed. For reference, see the [Express docs](https://expressjs.com/en/guide/using-middleware.html#middleware.router).

PR #89, opened 2018-07-23
fix(core): take devDependencies into account when resolving presets

robin-drexler:
Imo devDependencies should also be used to resolve presets.

For example, we are currently building a development proxy that is only needed for development. https://github.com/xing/hops/pull/522

jest-preset could be(come) another example.

Otherwise, you would either have to install devDependencies as dependencies or specify the presets manually.

### DIFF
--- a/packages/core/lib/config.js
+++ b/packages/core/lib/config.js
@@ -93,20 +93,26 @@ const loadConfig = (context, config) => {
     : searchSync(context);
 };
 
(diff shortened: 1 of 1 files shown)


robin-drexler:
@dmbch Also thought about only resolving devDependencies for non-prod envs and since you also came up with the idea, I added it. :D

PR #95, opened 2018-07-31
RFC: Introduce dynamic config

dmbch:


robin-drexler:
Like the direction that is going, although I'm not entirely sure if for testing purposes extending mixins is enough.

dmbch:
I am not quite sure I am getting your point here: you can essentially override every setting using this mechanism..: https://github.com/untool/untool/commit/94cf3127ad35b43e4937729aae2e35688d20d32b#diff-78c1a3e4f882c9f6848e2467f6272d0fR184

robin-drexler:
It is perfectly possible I'm wrong,  but doesn't  getConfig concat mixins? https://github.com/untool/untool/blob/bb307493c963874b010d7f4237160d35ee0f5bf5/packages/core/lib/config.js#L20

dmbch:
Ah, now I see what you mean. You are right wrt mixins being concat’ed - but (for testing/mocking purposes), you could change the strategy of any mixin method to, for example, `override` and inject your own implementation. 

Or you could misuse the strategies mechanism to implement mocks; strategies are just functions, after all...

Would these approaches help with the use cases you have in mind? 

Other than that, we might consider accepting functions as our `override` and pass them the raw config hash, but I must admit I am not in love with this idea.

dmbch:
@robin-drexler PTAnotherL

PR #97, opened 2018-08-03
Better yargs

dmbch:


packages/yargs/README.md
```diff
@@ -29,11 +29,17 @@ Options:
   --help, -h  Show help                                               [boolean]
 
 $ un start
-foo info
-server listening at http://localhost:8080
+foo listening at http://localhost:8080
```
robin-drexler: untool does not log such information anymore, does it? :D

dmbch: `untool` does since recently, but `@untool/yargs` still does not.

PR #104, opened 2018-08-13
Better exports

dmbch:


packages/cli/index.js
```diff
@@ -12,19 +12,28 @@ const getRootDir = () => {
...
-} catch (error) {
-  if (error.message && error.message.startsWith("Can't resolve")) {
-    // eslint-disable-next-line no-console
-    console.error("Error: Can't find @untool/yargs \n");
-  } else {
-    // eslint-disable-next-line no-console
-    console.error(error.stack ? error.stack.toString() : error.toString());
-  }
-  process.exit(1);
+exports.configure = (config = { mixins: [__dirname] }, options) => ({
```
ZauberNerd: I'd like to have an actual function body here for readability..

ZauberNerd: Also: I'm not too happy with the name `configure` - but don't know a better suggestion at the moment.

packages/core/lib/core.js
```diff
@@ -7,12 +7,13 @@ const { getConfig } = require('./config');
...
 exports.Mixin = class Mixin {
-  constructor(config) {
+  constructor(config, options) {
     this.config = config;
+    this.options = options;
   }
 };
 
-exports.bootstrap = function bootstrap(overrides, ...args) {
+exports.bootstrap = function bootstrap(overrides = {}, options = {}) {
```
ZauberNerd: Unrelated to this PR, but the name `overrides` is unclear. How about `configOverrides`?

packages/express/index.js
```diff
@@ -1,27 +1,22 @@
...
 
-module.exports = {
-  get uri() {
-    return require('./lib/uri');
-  },
-  runServer(mode, options = {}) {
-    const core = require('@untool/core').bootstrap();
-    core.handleArguments(options);
-    return core.runServer(mode);
+const { bootstrap } = require('@untool/core');
```
ZauberNerd: Just a tiny remark, but how about writing this code (in this file and all other files that export stuff) in a static manner, that doesn't mutate objects. I.e. like this:
```javascript
const { bootstrap } = require('@untool/core');
const uri = require('./lib/uri');

const configure = (config, options) => {
  return { ... };
};

const { runServer, createServer, createRenderer, renderLocations } = configure();

module.exports = exports = { runServer, createServer, createRenderer, renderLocations, configure };
```

With this approach it is easier to see what exactly gets exported and also this is a lot closer to ES modules (once they're implemented in Node):
```javascript
import { bootstrap } from '@untool/core';
import uri from './lib/uri';

const configure = (config, options) => {
  return { ... };
};

const { runServer, createServer, createRenderer, renderLocations } = configure();

export { runServer, createServer, createRenderer, renderLocations, configure };
```

dmbch: PTAL

ZauberNerd:
Shouldn't this mixin also extend the `@untool/core` mixin and then accept `config` and `options` in the constructor and call `super` with them? Or just lose the constructor entirely as it isn't doing anything? https://github.com/untool/untool/blob/bd27b3fff70f7fcb85d87510db7b4d0c8d81c249/packages/cli/mixin.core.js#L8

dmbch:
@ZauberNerd I think the `untool` mixin should not inherit from `@untool/core`: `untool` is primarily meant to be installed globally, and we would have to jump through some hoops to reliably get a grip on `@untool/core`.

I have added the missing `options` argument, though.

ZauberNerd:
Code-wise LGTM - what about documentation though?

dmbch:
@ZauberNerd docs will follow soonish - the previous exports were undocumented as well.

PR #111, opened 2018-08-20
Improve serverjs handling

dmbch:


dmbch:
@robin-drexler I would like to think this through. For example, I would hate the idea of having to deploy `node_modules/.cache/untool` while excluding `node_modules`...

PR #115, opened 2018-08-22
Improve asset handling

dmbch:


packages/webpack/lib/middleware/assets.js
```diff
@@ -3,41 +3,32 @@
...
-        Object.keys(assetsByType).forEach((extension) => {
-          assetsByType[extension].push(
-            ...assets.filter(
-              (asset) =>
-                asset &&
-                extname(asset) === `.${extension}` &&
-                !asset.endsWith('.hot-update.js')
-            )
-          );
+        [].concat(assetsByChunkName[chunkName]).forEach((asset) => {
```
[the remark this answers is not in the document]
dmbch: Do you really think this is better readable..?

```javascript
(
  Array.isArray(assetsByChunkName[chunkName])
    ? assetsByChunkName[chunkName]
    : [assetsByChunkName[chunkName]]
)
```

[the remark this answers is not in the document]
ZauberNerd: How about something like this?
```javascript
Object.entries(assetsByChunkName)
  .filter(([name]) =>
    new RegExp(`^(vendors~)?${config.name}$`).test(name)
  )
  .sort(([name]) => (name.startsWith('vendors~') ? 0 : 1))
  .map(([name, assets]) => [name, Array.isArray(assets) ? assets : [assets]])
  .forEach(([, assets]) => {
    assets.forEach((asset) => {
      const extension = extname(asset).substring(1);
      if (assetsByType[extension] && !asset.endsWith('.hot-update.js')) {
        assetsByType[extension].push(asset);
      }
    });
  });
```

**Or:**
```javascript
const assetsByType = Object.entries(assetsByChunkName)
  .filter(([name]) =>
    new RegExp(`^(vendors~)?${config.name}$`).test(name)
  )
  .sort(([name]) => (name.startsWith('vendors~') ? 0 : 1))
  .map(([name, assets]) => [name, Array.isArray(assets) ? assets : [assets]])
  .reduce((assetsByType, [, assets]) => {
    assets.forEach((asset) => {
      const extension = extname(asset).substring(1);
      if (assetsByType[extension] && !asset.endsWith('.hot-update.js')) {
        assetsByType[extension].push(asset);
      }
    });
    return assetsByType;
  }, { css: [], js: [] });
```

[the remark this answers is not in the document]
dmbch: 👍

dmbch:
@ZauberNerd PTAnotherL

dmbch:
@ZauberNerd, @robin-drexler: I managed to streamline and somewhat simplify the async, state-machine-y stuff we have to do for Node rendering. I’d appreciate your input.

packages/webpack/mixin.core.js
```diff
@@ -11,17 +11,28 @@ const {
...
-    const { config, assets, assetsByChunkName } = this;
-    return assetsMiddleware(config, { assets, assetsByChunkName });
+  constructor(...args) {
+    super(...args);
+    this.assets = new Resolvable();
+  }
+  loadPrebuiltAssets() {
+    const { serverDir, assetFile } = this.config;
+    const path = join(serverDir, assetFile);
+    return exists(path) ? require(path) : {};
```
ZauberNerd: Should we print a warning in case the `assetFile` does not exist?

packages/webpack/lib/middleware/render.js
```diff
@@ -1,77 +1,40 @@
...
-  const transpiler = exports.createTranspiler(webpackConfig);
-  let middleware, error;
-  transpiler.on('start', () => {
-    middleware = null;
-    error = null;
-  });
-  transpiler.on('success', (result) => {
-    middleware = result;
-    error = null;
+  const resolvable = new Resolvable();
```
ZauberNerd: :+1: this module is also pretty clean now.

packages/webpack/lib/middleware/assets.js
```diff
@@ -1,43 +1,12 @@
...
-              (asset) =>
-                asset &&
-                extname(asset) === `.${extension}` &&
-                !asset.endsWith('.hot-update.js')
-            )
-          );
-        });
-      });
-    next();
+module.exports = (assets) => {
```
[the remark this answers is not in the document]
ZauberNerd: One thing though: Would it make sense to not pass the entire `assets` Resolvable to this factory function but instead only pass the bound `assets.then` method - as to indicate inside the core mixin, that this middleware will only read and not mutate assets?

packages/webpack/mixin.core.js
```diff
@@ -11,17 +11,28 @@ const {
...
   }
   loadPrebuiltMiddleware() {
     const { serverDir, serverFile } = this.config;
     const path = join(serverDir, serverFile);
     return exists(path) ? require(path) : (req, res, next) => next();
   }
+  createAssetsMiddleware() {
+    const assetsMiddleware = require('./lib/middleware/assets');
+    const { assets } = this;
+    return assetsMiddleware(assets);
```
ZauberNerd: As per my comment above in the assets middleware: Would it make sense to somehow indicate, that the assets middleware will use `this.assets` only to read instead of read and write to it?
Maybe something like this could already clear things up?
```javascript
return assetsMiddleware({ getAssets: this.assets.then.bind(this.assets) });
```
Or `resolvedAssets` instead of `getAssets` or something similar.

packages/webpack/lib/plugins/assets.js
```diff
@@ -1,34 +1,55 @@
...
       );
-      if (this.options.static) {
-        this.setAssets(assets, assetsByChunkName);
-      } else {
-        compilation.assets[this.config.assetFile] = new RawSource(
-          JSON.stringify({
-            assets: assets,
-            assetsByChunkName: assetsByChunkName,
-          })
+    } else {
```
ZauberNerd: So, if I understand correctly, this else branch will be executed for `un build` or each watch-run of `un develop` and will then read/reduce the assets from the current compilation and resolve the parent (mixin's) `assets` promise to their result, right?

And that result will then be used in the `if` branch above to write the `assetsFile` to the filesystem (in case of `un build`; or the memory fs, when running `un develop`).

packages/webpack/lib/middleware/render.js
```diff
@@ -1,77 +1,40 @@
...
-    emitter.emit(...args);
-    emitter.emit('result');
-  };
   const handleCompilation = (compileError, stats) => {
     if (compileError) {
-      emitResult('error', compileError);
+      resolvable.reject(compileError);
     } else if (stats.hasErrors()) {
-      stats.toJson().errors.forEach(emitResult.bind(null, 'error'));
+      stats.toJson().errors.forEach(resolvable.reject);
```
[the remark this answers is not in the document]
dmbch: With regards to multiple errors you are completely right. Since this is code that only ever runs in developemnt or during static builds I considered that acceptable. If we wanted to change that, we would have to build some middleware capable of handling multiple errors...

Your recollection is spot on: whatever error is being passed to `resolvable.reject()` here is being passed to Express' `next()` later on.

packages/webpack/lib/middleware/render.js
```diff
@@ -1,77 +1,40 @@
...
     } else if (stats.hasWarnings()) {
-      stats.toJson().warnings.forEach(emitResult.bind(null, 'error'));
+      stats.toJson().warnings.forEach(resolvable.reject);
```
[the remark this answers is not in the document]
dmbch: I think we do - at least we did so since the very beginning of hops/untool... Of course we could change that behaviour, but I am not sure I would do so in this PR.

Same answers to the other questions above - though I kinda think that the concurrency issue is a bit edgy...

packages/webpack/mixin.core.js
dmbch: I have to think this through - too many cases...

packages/webpack/lib/plugins/assets.js
dmbch: You *do* understand correctly 😃

packages/webpack/lib/middleware/assets.js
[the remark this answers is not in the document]
dmbch: To be honest, passing around a (bound) `then()` method feels kind of strange to me personally - even if we did rename it..: `getAssets(assets => {...}, error => {...})`

Additionally, from the perspective of the mixin, `assets` currently is a rather opaque object: the mixin only knows who needs to get it, but is not at all concerned with how it is being used. Does that make sense to you?

packages/webpack/mixin.core.js
dmbch: Please see above: https://github.com/untool/untool/pull/115#discussion_r213105035

packages/webpack/lib/middleware/assets.js
```diff
@@ -1,43 +1,12 @@
...
+  return function assetsMiddleware(req, res, next) {
+    assets
+      .then((assets) => {
+        res.locals = { ...(res.locals || {}), ...assets };
```
jhiode: How about this:
```res.locals = { ...res.locals, ...assets };```
That should work even if `res.locals` is `undefined` or `null`.

File: packages/webpack/lib/middleware/render.js [hunk 10]
```diff
@@ -1,77 +1,41 @@
...
-      emitResult('error', compileError);
-    } else if (stats.hasErrors()) {
-      stats.toJson().errors.forEach(emitResult.bind(null, 'error'));
-    } else if (stats.hasWarnings()) {
-      stats.toJson().warnings.forEach(emitResult.bind(null, 'error'));
+      resolvable.reject(compileError);
+    } else if (stats.hasErrors() || stats.hasWarnings()) {
+      const { errors, warnings } = stats.toJson();
+      resolvable.reject(
+        new Error(`Can't compile:\n${[].concat(errors, warnings).join('\n')}`)
```
jhiode: ```new Error([`Can't compile:`, ...errors, ...warnings].join('\n'))``` ? 😄

packages/webpack/lib/utils/resolvable.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+
+exports.Resolvable = class Resolvable {
```
jhiode: I like this approach, but I think this file could be much simpler as the promise class already has a queueing behaviour.

```
exports.Resolvable = class Resolvable {
  constructor() {
    this.reset = this.reset.bind(this);
    this.reset();
  }

  reset() {
    const promise = new Promise((resolve, reject) => {
      this.resolve = resolve;
      this.reject = reject;
    });

    this.then = promise.then.bind(promise);
    this.catch = promise.catch.bind(promise);
  }
};
```

I've ignored the `executor` logic as it is not used...

dmbch: Nice one!

Unfortunately, it leaves us with 'orphaned' listeners - imagine a scenario as follows:
* compilation starts (`rslvbl.reset()`)
* listener subscribes  (`rslvbl.then()`)
* compilation restarts (`rslvbl.reset()`)
* compilation finishes (`rslvbl.resolve('foo')`) 
* listener is still unresolved... 😒

How about this implementation (including `executor`):

```javascript
class Resolvable {
  constructor(executor) {
    this.reset = () => {
      const { resolve: oldResolve, reject: oldReject } = this;
      const promise = new Promise((newResolve, newReject) => {
        this.resolve = (value) => {
          if (oldResolve) oldResolve(value);
          newResolve(value);
        };
        this.reject = (reason) => {
          if (oldReject) oldReject(reason);
          newReject(reason);
        };
      });
      this.then = promise.then.bind(promise);
      this.catch = promise.catch.bind(promise);
    };
    this.reset();
    if (executor) {
      new Promise(executor).then(this.resolve, this.reject);
    }
  }
}
```

PR #121, opened 2018-08-29
Babel 7

dmbch:
Since Babel version 7.0.0 has finally been [released](https://babeljs.io/blog/2018/08/27/7.0.0), it is high time to upgrade. 🚀 

Closes: #119


dmbch:
@jhiode because I wanted to prevent being breaking (again) - the contents of that preset used to be part of the React preset, see https://babeljs.io/docs/en/babel-preset-react.html

That being said: that preset only removes Flow type annotations and does not do anything else, see https://babeljs.io/docs/en/babel-preset-flow/

Anyway, I would not really mind getting rid of it - what do you think?

jhiode:
I was just wondering what the reason for that was. Being non breaking sounds good 🙂

dmbch:
@robin-drexler I had to in order to get `@babel/preset-env` working with `useBuiltIns: 'usage'`. After changing almost everything runtime-y to ESM (and inlining `environmentalize`), I managed to get things working now. PTAL.

Issue #120, opened 2018-08-28
Do not crash the server on warnings


ZauberNerd:
Do not reject the build when webpack generates warnings (i.e. only reject on errors) for server-side builds.

PR #122, opened 2018-08-29
Make render and asset middlewares more robust

dmbch:
Closes: #120


packages/webpack/mixin.core.js
```diff
@@ -57,7 +57,6 @@ class WebpackMixin extends Mixin {
         logLevel: 'silent',
         publicPath: webpackBrowserConfig.output.publicPath,
         watchOptions: webpackBrowserConfig.watchOptions,
-        serverSideRender: true,
```
ZauberNerd: What does this change do?

packages/webpack/lib/middleware/render.js
```diff
@@ -1,41 +1,79 @@
...
-    } else if (stats.hasErrors() || stats.hasWarnings()) {
-      const { errors, warnings } = stats.toJson();
-      resolvable.reject(
-        new Error(`Can't compile:\n${[...errors, ...warnings].join('\n')}`)
-      );
+      emitResult('error', compileError);
+    } else if (stats.hasErrors()) {
+      const { errors } = stats.toJson();
+      const details = errors
+        .map((str) => str.replace(/\033?\[[0-9]{1,2}m/g, ''))
```
ZauberNerd: What are you doing here? Replacing color codes?

ZauberNerd: Seems like you can set `colors: false` to the `stats.toJson()` method - https://webpack.js.org/configuration/stats/

dmbch: Yes, stripping color codes - I tried just using that setting, but it does not work the way we both thought. ;)

packages/webpack/mixin.core.js
dmbch: That setting instructs `webpack-dev-middleware` to make sure it sets the stats to `res.locals.webpackStats` before invoking the next middleware, allowing a developer to render the page body and manage the response to clients. See: https://github.com/webpack/webpack-dev-middleware#server-side-rendering

We do not use that feature any more - and it is marked as `experimental`, anyway.

packages/webpack/lib/utils/resolvable.js
```diff
@@ -17,28 +19,33 @@ exports.Resolvable = class Resolvable {
...
+            (error) => void reject(error) || this.reject(error),
+            (executor) => this.reset(executor)
+          )
+        ).catch(this.reject);
+      }
+    };
+    this.resolve = (value) => {
+      Promise.resolve(value).then(
+        (value) => queue.slice().forEach((fn) => fn(null, value)),
+        this.reject
```
jhiode: Why do we need this additional promise? And this `reject` function will never be called as the promise is already resolved.

dmbch: I am afraid we do: `value` could be a `Promise`. That is the reason I added `this.reset` here, too.

Generally, I want this thing to mimick the behavior of a real `Promise`.

PR #126, opened 2018-09-06
Better source maps

dmbch:
To enable source-map support for node *elegantly* in both dev and prod environments, we need to write the transpiled code to disk in all cases. This PR ensures we do just that - it might break things downstream, though, if someone were to rely upon the fact that we used to not to.

However, I still think this is the right thing to do, as it enables us to get rid of both `MemoryFS` and `requireFromString`, both of which used to add considerable complexity. Oh, and we do not have to rely on `source-map-support`'s `hookRequire` functionality any more.


dmbch:
I’ll give this a once-over today. Will re-open if required.

PR #137, opened 2018-09-13
fix(webpack): only call configureBuild once for develop client bundle

robin-drexler:
Coming from https://github.com/xing/hops/issues/636

When target is `develop`, `configureBuild` is executed twice for the client bundle.
That is not the case for target `build`.


dmbch:
Thanks for this PR! I would love to understand the problem we are addressing with it, though.

This actually is the expected behavior: https://github.com/untool/untool/blob/master/packages/webpack/mixin.core.js#L40

If this is an actual problem, your solution is certainly valid and I’d be happy to merge it.

PR #147, opened 2018-09-17
Introduce universal React code splitting

dmbch:
```javascript
import React from 'react';
import { Switch, Route, Link } from 'react-router-dom';

import { render } from '@untool/core';
import { Import, Miss } from '@untool/react';

const Home = Import('./home');
const About = Import('./about', 'About');

const preload = ({ Component, error, loading, ...props }) => {
  if (loading) return <p>Loading...</p>;
  if (error) return <p>Error!!!</p>;
  return <Component {...props} />;
};

const App = () => (
  <>
    <Link to="/">Home</Link> <Link to="/about">About</Link>
    <Switch>
      <Route exact path="/" component={Home} />} />
      <Route path="/about" render={() => <About render={preload} />} />
      <Miss />
    </Switch>
  </>
);

export default render(<App />);
```


packages/react/lib/components.js
```diff
@@ -0,0 +1,55 @@
...
+});
+
+exports.Import = ({ loader, loading, weakId, module }) =>
+  withRouter(
+    class ImportComponent extends PureComponent {
+      constructor({ staticContext }) {
+        super();
+        if (__webpack_modules__[weakId] || staticContext) {
+          this.state = {
+            Component: ((value) => value.default || value)(
```
ZauberNerd: This won't work with components that use named exports.

tests/fixtures/untest/index.js
```diff
@@ -1,5 +1,26 @@
...
-import React from 'react';
+/* eslint-disable no-unused-vars */
+import React, { Fragment } from 'react';
+import { Helmet } from 'react-helmet';
+import { Switch, Route, Link } from 'react-router-dom/es';
+
 import { render } from '@untool/core';
+import { Import, Miss } from '@untool/react';
+
+const Home = Import('./home');
```
ZauberNerd: I'm not a big fan of this API..
- it's too easy to oversee that the first letter is a capital `I`
- it doesn't allow to customize rendering
  - rendering of a loading screen
  - rendering of a component that uses named exports
  - setting a delay for the 'loading..' screen (to avoid flashes on fast network connections / caches)
  - rendering alternative / fallback content
- it doesn't allow to customize loading
  - specifying network timeouts
  - retrying failed requests
  - loading more than one component / loading data (ie. `Promise.all(...)`)

packages/react/lib/babel.js
```diff
@@ -0,0 +1,54 @@
+'use strict';
+
+module.exports = function({ types: t }) {
+  return {
+    visitor: {
+      CallExpression(path) {
+        if (path.node.callee.name !== 'Import') return;
```
ZauberNerd: This is too fragile - we should check for the import path, to catch things where the function is being imported under a different name:
- `import { Import as LoadComponent } from '@untool/react';`
- or `const { Import: LoadComponent } = require('@untool/react');`
- or `const LoadComponent = require('@untool/react').Import;`

packages/react/lib/components.js
dmbch: Yeah, that is true. I would argue we can tackle this later iff we need to - either here or downstream. react-loadable does so by supporting and optional `render` function:

https://github.com/jamiebuilds/react-loadable/blob/master/README.md#customizing-rendering

tests/fixtures/untest/index.js
dmbch: Let us agree to disagree about `{i|I}mport` - I really like it... 😉 

With regards to loading/rendering, please see [above](https://github.com/untool/untool/pull/147#discussion_r219474414).

packages/react/lib/babel.js
dmbch: I am not overly happy about it neither - do you have an idea how to find out the name of an imported/required function? Babel's docs are still a bit sparse...

packages/react/lib/components.js
dmbch: I have tackled this now by adding a second (optional) argument to `Import(module [, name])`. Of course, it defaults to `default` :grin:

tests/fixtures/untest/index.js
dmbch: You can now add a `placeholder` prop to components created with `Import()`: this will allow implementers to build whatever loading logic and UI they fancy.

packages/react/lib/babel.js
```diff
@@ -0,0 +1,52 @@
+'use strict';
+
+module.exports = function({ types: t }) {
+  return {
+    visitor: {
+      ImportDeclaration(path) {
+        const source = path.node.source.value;
+        if (source !== '@untool/react') return;
```
ZauberNerd: This should be configurable.

packages/react/lib/babel.js
```diff
@@ -0,0 +1,52 @@
...
+
+        const specifiers = path.get('specifiers');
+        const specifier = specifiers.find(
+          (specifier) => specifier.node.imported.name === 'Import'
```
ZauberNerd: This doesn't need to be, but could also be configurable..

packages/react/lib/components.js
```diff
@@ -0,0 +1,63 @@
...
+      return render({ ...realProps, ...this.state });
+    }
+  }
+  return function ImportWrapper({ placeholder, render, ...realProps }) {
+    return createElement(
+      withRouter(function ImportManager({ staticContext }) {
+        if (staticContext) {
+          staticContext.modules.push(module);
+        }
+        if (staticContext || __webpack_modules__[weakId]) {
```
ZauberNerd: How about having another variable, to clear up any confusion about why you are checking `staticContext` here?
Something like: `const isNode = typeof staticContext !== 'undefined';`
Or maybe add a comment, describing why you are testing for `staticContext` or `__webpack_modules__[weakId]`.

ZauberNerd: Also: Why are you testing for `__webpack_modules__[weakId]` and are using `__webpack_require__(weakId)` in the next line?

packages/react/lib/components.js
```diff
@@ -0,0 +1,63 @@
...
+    }
+    render() {
+      const defaultRender = ({ Component, error, loading, ...props }) => {
+        return !(error || loading) ? createElement(Component, props) : null;
+      };
+      const { render = defaultRender, realProps } = this.props;
+      return render({ ...realProps, ...this.state });
+    }
+  }
+  return function ImportWrapper({ placeholder, render, ...realProps }) {
```
ZauberNerd: How about `ownProps` instead of `realProps`? At least that's the name that is used in some other libraries (eg. redux).

packages/react/lib/components.js
```diff
@@ -0,0 +1,63 @@
...
+      withRouter(function ImportManager({ staticContext }) {
+        if (staticContext) {
+          staticContext.modules.push(module);
+        }
+        if (staticContext || __webpack_modules__[weakId]) {
+          const Component = __webpack_require__(weakId)[name];
+          return createElement(Component, realProps);
+        } else {
+          const Placeholder = placeholder || ImportPlaceholder;
+          return createElement(Placeholder, { realProps, render, load, name });
```
ZauberNerd: `name` is pretty generic - how about `exportSpecifier` or something similar?

packages/react/lib/components.js
```diff
@@ -0,0 +1,63 @@
...
+exports.Header = withRouter(({ staticContext, name, value }) => {
+  if (staticContext) {
+    staticContext.headers = { ...staticContext.headers, [name]: value };
+  }
+  return null;
+});
+
+exports.Import = ({ module, load, weakId }, name = 'default') => {
+  class ImportPlaceholder extends Component {
+    constructor({ load, name }) {
```
ZauberNerd: Why use the constructor and not `componentWillMount` or `componentDidMount`? If I understand this correctly, the `ImportPlaceholder` will never be used on the server anyways, right?

packages/react/lib/components.js
dmbch: Fair point. We could go with `exportName` here - the things are called “named exports”, after all.

PR #150, opened 2018-09-20
[RFC]: feat(webpack): pass "target" to webpack config loader

ZauberNerd:
This change will pass the "target" argument (can be one of "browser" or
"server") to the config loader, to allow to create a webpack
config for a specific mixin type. It also adds three new hooks, namely
"getBuildConfig", "getDevelopConfig" and "getNodeConfig", to get the
webpack config for the corresponding target.


packages/webpack/mixin.core.js
```diff
@@ -63,11 +63,40 @@ class WebpackMixin extends Mixin {
     const { StatsFilePlugin } = require('./lib/plugins/stats');
     return new StatsFilePlugin(this);
   }
-  getBuildConfig(target) {
+  _oldGetBuildConfig(target) {
+    console.warn('getBuildConfig with target:', target, 'is deprecated.');
```
ZauberNerd: We should consider using https://www.npmjs.com/package/depd for this and have a `deprecations.md` file in the repository root where we describe the deprecations in detail: https://github.com/nodejs/node/blob/master/doc/api/deprecations.md

packages/webpack/index.js
```diff
@@ -12,6 +12,12 @@ const configure = (config, options) => ({
   getBuildConfig(...args) {
     return bootstrap(config, options).getBuildConfig(...args);
   },
+  getDevelopConfig(...args) {
```
ZauberNerd: Maybe we should consider renaming these to `getWebpack*Config` instead, to make it more clear, what these hooks do.

packages/webpack/mixin.core.js
```diff
@@ -219,6 +232,8 @@ WebpackMixin.strategies = {
   configureBuild: pipe,
   inspectBuild: sequence,
   getBuildConfig: callableSync,
```
dmbch: These strategy definitions need to be rename to `getWebpack{Build,Develop,Node}Config`, respectively.

PR #153, opened 2018-09-25
Parameterize webpackConfig of dev/render middlewares

ZauberNerd:
In order to pass modified webpack configs to the dev- and render-
middlewares, we need to parameterize them.


packages/webpack/mixin.core.js
```diff
@@ -29,17 +29,13 @@ class WebpackMixin extends Mixin {
       return (req, res, next) => next();
     }
   }
-  createRenderMiddleware() {
-    const createRenderMiddleware = require('./lib/middleware/render');
-    const webpackConfig = this.getWebpackNodeConfig();
-    return createRenderMiddleware(webpackConfig);
+  createRenderMiddleware(webpackConfig) {
+    return require('./lib/middleware/render')(webpackConfig);
```
ZauberNerd: Now we could also get rid of this class method, because it doesn't do anything anymore.

ZauberNerd: Also: I'd propose to export the render and stats middleware and plugins through `@untool/webpack`s index.js file, so that we can use them in other packages (we need the renderMiddleware and the StatsPlugin in Hops GraphQL).

And then we could just import the middlewares and plugins at the top of this file (`const { createRenderMiddleware, createStatsMiddleware, RenderPlugin, StatsPlugin } = require('@untool/webpack');`) and call / instantiate them directly wherever they are being used instead of wrapping them in one-line class methods that don't do anything extra.

PR #169, opened 2018-10-03
Prepare v1.0

dmbch:
Let's move on to version 1.0. In this, I changed the strategies applied by most non-`callable` `core` hooks to `sequence` to get rid of the necessity to, for example, alway return `webpackConfig`. This way, I think we will get rid of one of the remaining sharper edges.

Apart from that, I moved some things to where they 'naturally' belong: most importantly, the main `render` export was moved to `@untool/react`. Apart from that, `locations`, `{asset|base}Path` and related settings were moved to `@untool/webpack`; static rendering was cleaned up and mostly moved there, too. This is not quite what I am proposing in #170, but makes things more bearable in my opinion.

Additionally, I re-introduce an `untool` package to provide a single entry point for all other packages. It does not just provide a (minimal) CLI, but also exports for native node (`core`) and runtime (`browser` and `server`) environments.


dmbch:
Too much change, too little gain.

PR #195, opened 2018-10-23
Trying out a better fixture solution

dmbch:


dmbch:
This change must somehow introduce a weird side-effect I cannot seem to localise - closing for now... 🤷‍♂️

PR #206, opened 2018-10-31
feat(webpack): allow alias definition

### LINKED ISSUE #205: Allow webpack alias definition
I want to be able to define alias like

````javascript
const path = require('path');

module.exports = {
  "browsers": "last 1 Chrome versions",
  "alias": {
    "@": path.resolve(__dirname, 'src')
  },
  "locations": [
    "/",
    "/counter"
  ]
};
````
So, for example, we will be able to replace `../../../../../../../../components/Unicorn` by `@/components/Unicorn`

aach:
Fixes  #205

### DIFF
--- a/packages/webpack/lib/configs/build.js
+++ b/packages/webpack/lib/configs/build.js
@@ -85,9 +85,9 @@ module.exports = function getConfig(config, name) {
         relative(config.rootDir, info.absoluteResourcePath),
     },
     resolve: {
-      alias: {
+      alias: Object.assign({}, config.alias, {
         '@untool/entrypoint': config.rootDir,
-      },
+      }),
       extensions: ['.mjs', '.js'],
       mainFields: [
         'esnext:browser',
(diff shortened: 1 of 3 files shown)


dmbch:
Hej @aach, thank you for this PR! The topic of relative imports keeps coming up - until now, we have refrained from explicitly adding support using a mechanism similar to your proposal. In such cases, I have often recommended using [`yarn workspaces`](https://yarnpkg.com/lang/en/docs/workspaces/) as an alternative, more low-level solution that also works for plain Node.js, test runners such as Jest or Mocha and for editors/IDEs.

If workspaces do not work for your project, would an alternative implementation using a `NODE_PATH` environment variable instead of an explicit `untool` config do? This would be similar to what [`create-react-app`](https://github.com/facebook/create-react-app/blob/4562ab6fd80c3e18858b3a9a3828810944c35e36/packages/react-scripts/config/env.js#L51-L65) does - and I kinda like that approach as it mimics [Node.js](https://nodejs.org/api/modules.html#modules_loading_from_the_global_folders)' behaviour.

aach:
Thanks for your response @dmbch! I will check your proposal and come back :)

aach:
Sorry for the delay! I have created #214 that accepts `NODE_PATH`

PR #207, opened 2018-11-05
Remove non-hook mixin methods from `@untool/react`

dmbch:
We decided to strongly discourage adding non-hook methods to mixins as doing so may result in namespace clashes. In this PR, I added appropriate documentation and refactored `@untool/react`'s server mixin to make sure we ourselves adhere to this new best practice.

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -163,7 +163,9 @@ MyMixin.strategies = {
 export default MyMixin;
 ```
 
-If inheriting from `Mixin`, all methods of your mixin are automatically bound to the repective instance, so you do not have to call `this.method.bind(this)` yourself even if you use them in asynchronous contexts.
+If inheriting from `Mixin`, all mixinable methods of your mixin are automatically bound to the respective instance, so you do not have to call `method.bind()` yourself even if you use them in asynchronous contexts.
+
+While it is technically possible to define non-mixin utility methods on your mixin, doing so is strongly discouraged. If you have to, however, it is recommended to prefix such methods' names with an underscore (`_`) to denote them as private.
 
 Note that you can call all defined mixinable methods directly on your mixin instance.
 
--- a/packages/react/lib/render.js
+++ b/packages/react/lib/render.js
@@ -1,0 +1,46 @@
+'use strict';
+
+const { extname } = require('path');
+
+const { renderToString } = require('react-dom/server');
+const { Helmet } = require('react-helmet');
+
+function determineAssets(modules, stats) {
+  const { entryFiles, vendorFiles, moduleFileMap } = stats;
+  const moduleFiles = modules.reduce(
+    (result, module) => [...result, ...moduleFileMap[module]],
+    []
+  );
+  return [
+    ...vendorFiles.sort((a, b) => b.localeCompare(a)),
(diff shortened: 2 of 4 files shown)


ZauberNerd:
@dmbch I don't see how this avoids name clashes at all. With this we're not fixing the issue but just closing our eyes for it.

A different approach that I've been thinking about might be:
- each mixin class needs to opt-in its methods that should be visible to mixinable
  ```
  MyMixin1 extends Mixin {
    handleError() {
      // this function opts-in to being called from the mixin
      // but will be overwritten by MyMixin2
    }
  }
  MyMixin1.strategies = {
    handleError: strategies.overwrite,
  };

  MyMixin2 extends Mixin {
    handleError() {
      // this function will actually be executed
      // when mixin.handleError() is being called
    }
  }
  MyMixin2.strategies = {
    handleError: strategies.overwrite,
  };

  MyMixin3 extends Mixin {
    handleError() {
      // this method will not be visible to the mixin
      // and therefore won't be called or overwritten
    }
  }
  ```
- the mixin methods should not be directly mixed-in onto `this` but instead should be made available as another property on each instance (eg. `this.mixin`) to allow classes to call own methods without fearing to interfere / call other mixin methods.
- we then must validate that all declared strategies use the same strategy implementation (eg. throw if multiple mixins opt-in their `handleError` method, but some define it as `overwrite` and others as `parallel`.
- or we could collect all methods per mixin strategy and apply each with their defined strategy?

I haven't thought this fully through and it definitely needs more investation and maybe we can't ship this with Hops 11, but should we take some more time to investigate this? What are your thoughts?

/cc @dmbch @robin-drexler @herschel666 @jhiode

dmbch:
@ZauberNerd that seems like an interesting change that we might want to consider for `untool/mixinable`. That being said, I am not quite convinced, yet.

* having multiple mixin strategy definition does not prevent namespace clashes, either; e.g. there might be multiple `handleError` implementations with different intended use cases (or signatures) and we would still be unable to distinguish between them
* I disagree with your assessment of the validity of this approach that basically implies that mixins ought to only contain methods intended to be used as mixinable hook implementations; this does not prevent namespace clashes, but at least prevents inadvertent hook implementation, i.e. the issue I actually intended to tackle here; I should have made that a lot clearer in the PR description
* I am not really sure if not overriding mixinstance methods with mixinable methods helps at all in this context: paradigmatically, we _are_ really building complex composite object here, after all

We could, of course, implement a pure opt-in solution with mixins (optionally) exposing an `implementations` array and only use methods whose names are included in said array. Such an approach might prevent inadvertent mixin implementation, as well.

Additionally, we might want to add a sanity check to `@untool/core` to make sure we throw an exception if strategies are being redefined. This would, of course, prevent some (potentially hacky) use cases (I am not quite sure we really want to support).

dmbch:
We could also consider switching to [`Tapable`](https://github.com/webpack/tapable) instead of `mixinable`, at least for core mixins/plugins. Doing so would make things a lot more explicit, of course.

I am not quite convinced this would be feasible in case of runtime mixins, though - using tapable's [`Context`](https://github.com/webpack/tapable#context) feature to implement the equivalents of `pipe`/`compose` strategies seems a bit unwieldy.

PR #208, opened 2018-11-05
RFC: Use static mixin properties as strategies

dmbch:
I was wondering if we really need an explicit `strategies` property - this is what it would take to get rid of it. What do you think, @untool/untool?


herschel666:
I'm not really convinced. I think it's worth the effort to have an explicit `strategies`-object.

dmbch:
Having thought this through once or twice more, I have come to agree with @herschel666.

PR #209, opened 2018-11-08
perf(webpack): only re-evaluate config when overrides exist

ZauberNerd:
We noticed that running "placeholdify" etc on the config is extremely
costly and at the moment we were executing it on every request.

With this change we will only re-evaluate the config whenever
"overrides" are actually used (most of the times it is just an empty
object).

We measured reqs/sec and on my machine (with hops-template-react), we
got on average:
- before: ~1700 reqs/sec
- after: ~2100 reqs/sec

Co-authored-by: Robin Drexler <drexler.robin@gmail.com>


dmbch:
Brilliant, thanks!

PR #211, opened 2018-11-18
fix(webpack): prevent dev server crashing on error

dmbch:
Currently, in `develop` mode, the server needs to be manually restarted after a syntax error. It does not recover upon fixing the error, but rather outputs..:

```
-----8<-----
[HMR] You need to restart the application!
[HMR] unexpected require(../../../packages/react/mixins/mixin.server.js) from disposed module ../../../packages/core/lib/config.js
[HMR] unexpected require(util) to disposed module
----->8-----
```

This PR fixes that issue by switching the `server.js` shim back to ES5 to be able to re-require our main entrypoint upon update.


packages/webpack/lib/middlewares/render.js
```diff
@@ -4,61 +4,70 @@ const { join } = require('path');
...
-          const { outputFileSystem: fs } = compiler;
-          fs.readFile(filePath, 'utf-8', (readError, fileContents) => {
-            if (readError) return reject(readError);
-            resolve(requireFromString(fileContents, filePath).default);
-          });
+          const { path, filename } = webpackConfig.output;
+          const filePath = join(path, filename);
+          middleware = require(filePath);
         }
+        process.nextTick(() => resolve(middleware));
```
ZauberNerd: I just tested this and it turns out that `process.nextTick()` is actually not required :see_no_evil: 

```suggestion
        resolve(middleware);
```

PR #212, opened 2018-11-19
[WIP][RFC] perf(express): use node cluster module to speed up request handling

robin-drexler:
Just toyed with clusters a bit, but thought I'd still share.

On my old-ish early 2015 macbook with 4 core i5, I could go from ~2k requests/second to about 3.2k requests/second.

For some reason using `worker.send` right after forking did not work. Same for using `cluster.on('online')` Only sending a message from the worker first seemed to resolve the issue. 

So if anybody can help me with that I'm doing wrong (regardless of if we ever want something like that in untool) I'd be really grateful. :) Maybe it has something todo with waiting until the `getPort` promise resolves? 

I'm not sure how often people actually use multi-core machines in production, though.

### DIFF
--- a/packages/express/lib/getPort.js
+++ b/packages/express/lib/getPort.js
@@ -1,0 +1,7 @@
+const portfinder = require('portfinder');
+
+module.exports = function getPort(port) {
+  if (port) return Promise.resolve(port);
+  portfinder.basePort = 8080;
+  return portfinder.getPortPromise();
+};
--- a/packages/express/lib/run.js
+++ b/packages/express/lib/run.js
@@ -5,8 +5,7 @@ const { join } = require('path');
 
 const { createServer: createHTTPServer } = require('http');
 const { createServer: createHTTPSServer } = require('https');
-
-const portfinder = require('portfinder');
+const getPort = require('./getPort');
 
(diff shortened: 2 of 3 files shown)


dmbch:
@robin-drexler I tried to reproduce the issues you were describing w.r.t. sending messages from master to workers - unfortunately I could not..:

```javascript
var cluster = require('cluster');

if (cluster.isWorker) {
  process.on('message', function(msg) {
    console.log(msg);
  });
}

if (cluster.isMaster) {
  for (var i = 0; i < 2; i++) {
    const worker = cluster.fork();
    worker.send('Workers of the world, unite!');
  }
}
```

This yielded the following CLI output:

```text
$ node workers.js 
Workers of the world, unite!
Workers of the world, unite!
```

I really do not see why something similar to the above examples would not work in your port determination example. Maybe we should just take a look together tomorrow i.r.l.

That being said, I share your doubts about adding this feature to `untool`: I think I would rather introduce some mechanism to facilitate usage of PM2 or similar with `untool`-powered apps.

dmbch:
Let us revisit this iff we have a substantial need for improved performance - i.e. post-1.0 😁

PR #217, opened 2018-11-30
fix(react): ensure Symbol-polyfill is present in time

herschel666:
...which is a delicate issue since babel-preset-env does not parse React files for usages.

@dmbch PTAL

### DIFF
--- a/packages/react/mixins/mixin.core.js
+++ b/packages/react/mixins/mixin.core.js
@@ -8,6 +8,10 @@ module.exports = class ReactMixin extends Mixin {
     fileLoaderConfig.exclude.push(/\.jsx$/);
     jsLoaderConfig.test.push(/\.jsx$/);
 
+    webpackConfig.entry = [require.resolve('core-js/es6/symbol')].concat(
+      webpackConfig.entry
+    );
+
     jsLoaderConfig.exclude.push(
       /node_modules\/react-helmet/,
       /node_modules\/react-dom/,


robin-drexler:
This among other things fixes untool apps in ie11.

dmbch:
Why do we not just remove `react` (or `react-dom` - I am not quite sure where the issue  is originating) from `jsLoader`’s excludes? If it is actually `react` itself, there should be no significant performance implications...

dmbch:
Speaking of which: are we still convinced we should exclude any packages from transpilation at all? IIRC, it only gives us a perceivable performance boost for the initial transpilation as we have caching enabled.

robin-drexler:
@dmbch iirc you were the who found out that those three packages have a significant impact on transpilation and hence build time. If that's still the case, manually adding the Symbol polyfill could be a feasible tradeoff.

> it only gives us a perceivable performance boost for the initial transpilation as we have caching enabled.

Imo there's also value in having a faster initial transpilation, although I obviously prefer working software over a slightly faster build. 

That being said, I think I need to educate myself more about webpack and babel caches actually work especially when they are busted etc.

dmbch:
@robin-drexler yes, I am afraid this one was up to me... On a clean working copy, I just checked the different build times with and without the babel excludes:

**Baseline (w/ excludes)**

```text
$ cd tests/fixtures/untest; un build -p
...
Child build:
    Time: 8147ms
    ...
Child node:
    Time: 4514ms
    ...
✨  Done in 9.58s.
$ cd tests/fixtures/untest; un build -p
...
Child build:
    Time: 2960ms
    ...
Child node:
    Time: 2417ms
    ...
✨  Done in 4.10s.
```

**De-optimized/fixed (w/o excludes)**

```text
$ cd tests/fixtures/untest; un build -p
...
Child build:
    Time: 10530ms
    ...
Child node:
    Time: 6585ms
    ...
✨  Done in 11.99s.
$ cd tests/fixtures/untest; un build -p
...
Child build:
    Time: 3450ms
    ...
Child node:
    Time: 2938ms
    ...
✨  Done in 4.57s.
```

I did notice *some* slowdown of incremental rebuilds (HMR) - around 20-50ms per rebuild. All these numbers are completely unscientific, of course.

Nevertheless: while I am convinced we would be able to shave off a couple of seconds from our build times by keeping those `excludes`, I tend to give up trying and working around their adverse effects. The effect we are tackling here is really quite obscure and rather hard to detect. Playing whac-a-mole with such issues is not something I would like us to spend time with...

dmbch:
Btw., thank you @herschel666 and @jhiode for debugging the issue and opening this PR!

robin-drexler:
>  yes, I am afraid this one was up to me...

I didn't mean to put any blame on you. Actually, I think that was a great finding! 
Nobody at the time did expect this performance workaround to cause such troubles!

> The effect we are tackling here is really quite obscure and rather hard to detect. Playing whac-a-mole with such issues is not something I would like us to spend time on...

100% agree. I can tell it took some time going from "This is broken in IE. Seems to be the Symbol issue again. Didn't we fix this?" to "This is only broken in production but works in development" to "Wait, this actually never worked in prod, only dev" to  "It also and only breaks in dev when we remove the hmr middleware entry" to "Why is react.production.min.js not going through transpilation?" to "OMG! WTF! LET'S GO BOOZE and flip some tables on the way  out." :D

PR #227, opened 2018-12-10
[WIP] feat(webpack): add @babel/runtime for smaller bundle size

robin-drexler:
https://babeljs.io/docs/en/babel-runtime

Couldn't see any improvement in the small untool fixture app, but in a larger one it seemed to make an impact.
Let's call that app `crate` :D 
Here's the difference of js files when using `build -p`. Yes that's over 600k, but that's another story.

```
 624K -> 570K 
 55K  -> 40K
 40K  -> 30K
 28K  -> 27K
 26K  -> 26K
 26K  -> 26K
 26K  -> 26K
 25K  -> 25K
 25K  -> 25K
 22K  -> 22K
 21K  -> 21K
 19K  -> 19K
 18K  -> 17K
 17K  -> 17K
 16K  -> 16K
 16K  -> 15K
 15K  -> 15K
 15K  -> 15K
 15K  -> 15K
 78B  -> 78B
```


robin-drexler:
Hm, tests seem to fail. I'd take a closer look if that's an approach we'd like to take. 
I can imagine that you, @dmbch, already have thought about it and maybe decided against doing it for some reason I don't know yet.

dmbch:
I am all for doing this (if we manage to make this work properly). We should probably explore the transform’s options (particularly `corejs` and `useESModules`).

Generally, we will probably only have to add this optimization selectively, i.e. only for browser or prod builds.

Oh, and we should probably check if this change also affects gzip’ed bundle size: redundant code might just be gzip’ed away, after all. I optimistic this will be an improvement, but compression keeps surprising me.

robin-drexler:
@dmbch Turns out transpiling the babel code isn't the best idea. The runtime deps seemed to be replaced with themselves which didn't work too well. :D

dmbch:
@robin-drexler I am really not quite sure we are on the right track here: during some quick research, I stumbled over this issue/comment: https://github.com/babel/babel/issues/6629#issuecomment-416986687

Afterwards, I did a quick check with a rather progressive browser matrix: `["last 1 Chrome versions"]`. Unfortunately, using this, build output was quite a bit larger than with just `@babel/preset-env`. So, in my opinion, we should wait until https://github.com/babel/babel/issues/6629 is resolved. What do you say?

robin-drexler:
To me it's not really clear yet why the bundle size increases tbh. If I understand the [docs](https://babeljs.io/docs/en/babel-plugin-transform-runtime#corejs) correctly, having `corejs` set to false (which is the default) should not add core-js dependencies. 

But I did not look at the bundle content yet.

dmbch:
Mmmaybe its about `regenerator-runtime`..? Apparently, [`@babel/runtime`](https://babeljs.io/docs/en/babel-runtime) comes with it. I am just guessing here, though.

By the way, did you manage to compare gzip’ed sizes, yet?

robin-drexler:
> By the way, did you manage to compare gzip’ed sizes, yet?

No not yet. Yesterday I only fixed the self-transpiling issues.

dmbch:
@robin-drexler do you still see the need to follow up on this?

PR #228, opened 2018-12-11
Change default serverDir to buildDir

ZauberNerd:


dmbch:
This continues the discussion in https://github.com/xing/hops/pull/726#discussion_r239585775, I think. As already mentioned there, I would not be too keen on creating a sparse `dist` folder even in `development` mode.

Also, I would not want to change the internal config structure so soon before 1.0 - not without extremely convincing reasoning.

dmbch:
@ZauberNerd I just had another look at your proposal. By using a setting defined by `@untool/webpack` in the `@untool/express` preset, we are introducing a new dependency. We will at least have to add said dependency to the express preset’s `package.json`.

This would, of course, establish a circular dependency between the two. Generally, I would love to keep the possibility of using the Express stuff without Webpack at least with static builds.

PR #235, opened 2018-12-26
fix(webpack): simplify module/chunk mapping

### LINKED ISSUE #240: importComponent breaks in production with re-exported modules
This issue was opened in [Hops repo](https://github.com/xing/hops/issues/758), this is its follow-up.

[untool/react](https://github.com/untool/untool/blob/master/packages/react/lib/render.js#L11) iterates over a `moduleFileMap` object which contains a few **invalid keys** (`""` and `null`) when:

- Webpack's [**ModuleConcatenationPlugin**](https://webpack.js.org/plugins/module-concatenation-plugin/) is introduced in production mode by [untool/webpack](https://github.com/untool/untool/blob/7c9fa25a97804dc5b47aea4135719ac04daffb9d/packages/webpack/lib/configs/build.js#L132).
- User dynamically import re-exported modules

It seems to be a known [webpack semi-issue](https://github.com/webpack/webpack/issues/8062), in relation to which Webpack's maintainer suggested to use an alternative way to identify generated modules.

**Avoiding** to **spread non-iterables** entities [here](https://github.com/untool/untool/blob/7c9fa25a97804dc5b47aea4135719ac04daffb9d/packages/react/lib/render.js#L10) seems to fix the issue (but I couldn't guarantee about possible side-effects 😃):

```js
  const moduleFiles = modules.reduce(
    (result, module) => {
      const moduleFile = moduleFileMap[module] || [];
      return[...result, ...moduleFile]; 
    },
    []
  );
```

dmbch:
This *should* resolve xing/hops#758

The initial code was based on
https://github.com/jamiebuilds/react-loadable/blob/master/src/webpack.js#L18-L20

Resolves #240, too.


robin-drexler:
@dmbch can you give some information on why this change fixes the issue? I'm wondering what makes our requirements different from react-loadable's in this regard?

dmbch:
@robin-drexler to be honest, I literally just wrote down what @ZauberNerd proposed privately IRL. We will probably still have to investigate further and I am more than happy to do so (after my vacation, i.e. next week).

When using `webpack.optimize.ModuleConcatenationPlugin` (like we do for production builds), we have to deal with instances of `ConcatenatedModule` that {do|used to} not have a `rawRequest` property. Since @ZauberNerd managed to resolve the issue described in xing/hops#758, this appears to have changed in the meantime.

dmbch:
In addition to adding re-exports to the fixture in https://github.com/untool/untool/pull/235/commits/b25b1ff443dff8b50497f314a2b7280c10d9584b I investigated quite a bit further: at least in our setup instances of `ConcatenatedModule` do have an `id` field. We use that instead of the `rawRequest` property that react-loadable relies on.

@ZauberNerd's solution appears to be valid - imo, this one is good to go. I think we ought to merge and ship it. 🚢

toomuchdesign:
Fix issue #240 and produces the expected chunks map. 👍

Shall we add a reference to `react-loadable` source to explain what we did here?

PR #236, opened 2018-12-28
fix(webpack): ignore child-compilers in RenderPlugin

herschel666:
This should™ resolve xing/hops/issues/761.

### DIFF
--- a/packages/webpack/lib/plugins/render.js
+++ b/packages/webpack/lib/plugins/render.js
@@ -20,6 +20,9 @@ exports.RenderPlugin = class RenderPlugin {
         )
       );
       compiler.hooks.compilation.tap('RenderPlugin', (compilation) => {
+        if (compilation.compiler.isChild()) {
+          return;
+        }
         compilation.hooks.additionalAssets.tapPromise('RenderPlugin', () =>
           promise.then((results) =>
             results.forEach(({ outfile, content }) => {


robin-drexler:
Can you elaborate what hops-postcss does to run into this particular issue?

herschel666:
It adds a child compiler `'mini-css-extract-plugin'` to the compilation, so the `RenderPlugin` hooks into the `additionalAssets` twice.

packages/webpack/lib/plugins/render.js
```diff
@@ -20,6 +20,9 @@ exports.RenderPlugin = class RenderPlugin {
         )
       );
       compiler.hooks.compilation.tap('RenderPlugin', (compilation) => {
+        if (compilation.compiler.isChild()) {
```
dmbch: I did not know `Compiler` instances had an `isChild()` method - nice find!

Could we not also use it in the [`StatsPlugin`](https://github.com/untool/untool/blob/master/packages/webpack/lib/plugins/stats.js#L55)?

//cc @ZauberNerd

robin-drexler:
Can we merge and release this, together with #235 to prepare a new hops release?
We'll add a bunch of tests in hops to ensure stuff has actually been fixed there too. :)

dmbch:
This one is good to go, but I would love to have #235 investigated/understood a bit better to make sure we do not introduce a subtle regression somewhere.

herschel666:
failed

@dmbch Can you merge this? It doesn't work via Github UI for me & in the Terminal I am causing a merge commit, what doesn't seem to be appreciated (referring to the git history of the project…).

robin-drexler:
Merged on my machine™ @herschel666

PR #247, opened 2019-01-16
fix(react): do not error when importComponent module not found

robin-drexler:
When you `import` a component (or import a file that exports said component, even if you do not directly use that export), that you also import via `importComponent`, that component will end up in an entry or vendor bundle and thus not be available in the moduleFileMap.
Which causes the following error, when a page is loaded:

```
TypeError: moduleFileMap[module] is not iterable
```

You can reproduce this by going to `master` branch and placing:

```js
import { About as a } from './components/about';
```
in https://github.com/untool/untool/blob/master/tests/fixtures/untest/index.js#L6

This PR fixes the error by ignoring those modules.
I've also added a warning in dev, because this is a situation a user certainly does not want.
Maybe we even want to throw an actual error and don't let users do something like that? What do you think @dmbch ?

Btw: `header-max-length` is an utterly annoying linting rule. :D

### DIFF
--- a/packages/webpack/lib/plugins/stats.js
+++ b/packages/webpack/lib/plugins/stats.js
@@ -13,20 +13,16 @@ const analyzeCompilation = ({ chunks, chunkGroups }) => {
     ],
     []
   );
-  const chunksByModule = chunks
-    .filter(
-      (chunk) => !entryChunks.includes(chunk) && !vendorChunks.includes(chunk)
-    )
-    .reduce(
-      (result, chunk) =>
-        Array.from(chunk.modulesIterable).reduce((result, module) => {
-          const { chunks } = chunkGroups.find(({ chunks }) =>
-            chunks.includes(chunk)
-          );
-          return [...result, [module.id, chunks]];
-        }, result),
-      []
-    );
+  const chunksByModule = chunks.reduce(
+    (result, chunk) =>
+      Array.from(chunk.modulesIterable).reduce((result, module) => {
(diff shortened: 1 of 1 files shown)


packages/react/lib/render.js
```diff
@@ -7,10 +7,20 @@ const { Helmet } = require('react-helmet');
...
   const { entryFiles, vendorFiles, moduleFileMap } = stats;
-  const moduleFiles = modules.reduce(
-    (result, module) => [...result, ...moduleFileMap[module]],
-    []
-  );
+  const moduleFiles = modules.reduce((result, module) => {
+    if (!moduleFileMap[module]) {
+      if (process.env.NODE_ENV !== 'production') {
+        // eslint-disable-next-line no-console
+        console.warn(
```
dmbch: To be honest, I would prefer either silently swallowing this (as proposed by @toomuchdesign in #240) or throwing an error if we want to tackle this issue the way you propose.

If we wanted to introduce warnings for certain kinds of erroneous uses of `@untool`'s features, I would prefer us to introduce some sort of central mechanism (ahem, `hook`) instead of an ad-hoc implementation including `ENV` check.

robin-drexler: Fixing stats is probably the preferable option.
Thinking about it, if I'm not mistaken we can't reliably know if an affected module will eventually end up in an entry chunk or not (because it might be treeshaken later on in those cases where the imported module is only part of another file but not actually required)

dmbch:
@robin-drexler would you be so kind as to try and push only signed commits from now on? 💚

PR #248, opened 2019-01-16
feat(express): add support for res.locals

dmbch:


packages/webpack/lib/plugins/render.js
```diff
@@ -12,8 +12,9 @@ exports.RenderPlugin = class RenderPlugin {
         Promise.all(
           requests.map((request) =>
             render(request).then((content) => ({
-              outfile:
-                request.outfile || trimLeadingSlash(indexFile(request.url)),
+              outfile: trimLeadingSlash(
+                request.outfile || indexFile(request.url || request)
```
jhiode: Ok wait, what is the intention of this change: ` || request`?

dmbch: `request` can also be a string, see
https://github.com/untool/untool/blob/d0195333c07fa3725462cd924736dbd79f8a80fc/packages/express/lib/render.js#L9

PR #250, opened 2019-01-18
feat(express): add support for route/method handlers

dmbch:


robin-drexler:
@dmbch I think it'd be nice if the new API was [documented](https://github.com/untool/untool/tree/master/packages/express) too.

PR #254, opened 2019-01-22
feat(core): make mixin types configurable

dmbch:
Architecturally, this change makes `@untool/core` more extensible and prevents it from knowing anything about the browser/server distinction implemented using webpack.

Practically, this allows users to introduce custom mixin types (for example to implement extensible service worker presets) and thus build more advanced extensions.

### DIFF
--- a/packages/webpack/preset.js
+++ b/packages/webpack/preset.js
@@ -20,4 +20,14 @@ module.exports = {
     join(__dirname, 'mixins', 'start'),
     join(__dirname, 'mixins', 'stats'),
   ],
+  mixinTypes: {
+    browser: {
+      mainFiles: ['mixin.browser', 'mixin.runtime', 'mixin'],
+      mainFields: ['mixin:browser', 'mixin:runtime', 'mixin'],
+    },
+    server: {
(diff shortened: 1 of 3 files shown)


packages/webpack/preset.js
```diff
@@ -20,4 +20,14 @@ module.exports = {
     join(__dirname, 'mixins', 'start'),
     join(__dirname, 'mixins', 'stats'),
   ],
+  mixinTypes: {
+    browser: {
+      mainFiles: ['mixin.browser', 'mixin.runtime', 'mixin'],
+      mainFields: ['mixin:browser', 'mixin:runtime', 'mixin'],
```
jhiode: How about simplifying the config to something like this?
```
  mixinTypes: {
    browser: ['browser', 'runtime'],
    server: ['server', 'runtime'],
  },
```

dmbch: I thought about that, too, but ultimately decided to go for the full-on `enhanced-resolve` config here for two reasons:

- more flexibility for downstream (e.g. support for *.ts mixins)
- less indirection/magic (we just use what is being passed to us w/o any assumptions)

PR #260, opened 2019-01-30
Refactor loader

dmbch:
Use exported core config utils instead of relying on a deep require from `@untool/core`, cache all calls to `getConfig()` (not just those without overrides) and generally reformat/restructure code generation code.

//cc @skrings


packages/webpack/lib/utils/loader.js
```diff
@@ -15,25 +15,27 @@ const getMixins = (type, { _mixins: mixins }) => {
   return `[${requires.join(',')}]`;
 };
 
+const generateModuleCode = (type, config) => `
+${type === 'server' ? 'const { dirname, join } = require("path");' : ''}
```
herschel666: A single condition would be enough here imho.

```diff
- ${type === 'server' ? 'const { dirname, join } = require("path");' : ''}
- ${type === 'server' ? 'const { sync: findUp } = require("find-up");' : ''}
- ${type === 'server' ? 'const rootDir = dirname(findUp("package.json"));' : ''}
- ${type === 'server' ? 'const expand = join.bind(null, rootDir);' : ''}
+ ${type === 'server' ? `
+   const { dirname, join } = require("path");
+   const { sync: findUp } = require("find-up");
+   const rootDir = dirname(findUp("package.json"));
+   const expand = join.bind(null, rootDir);
+ ` : ''}
```

dmbch: Yes, I know - probably another matter of personal/aesthetic preference, I guess. I think an additional `const` might do the trick...

dmbch: Done - made it another helper function (technically also a `const`).

PR #261, opened 2019-02-02
Refactor render

dmbch:
This removes `markup` and `mountpoint` from the object passed to `getTemplateData`: the former was renamed and moved to `fragments` while the latter was effectively obsolete as it could only be overridden server-side.

Besides that, this change adds support for a webpack `publicPath` by removing hardcoded slashes from asset URLs.


packages/webpack/lib/plugins/stats.js
```diff
@@ -26,8 +28,13 @@ const analyzeCompilation = ({ chunks, chunkGroups }) => {
...
 };
 
-const extractFiles = ({ entryChunks, vendorChunks, chunksByModule }) => {
-  const gatherFiles = (result, { files }) => [...result, ...files];
+const extractFiles = (chunkData, rawPublicPath) => {
+  const publicPath = trimTrailingSlash(rawPublicPath);
+  const { entryChunks, vendorChunks, chunksByModule } = chunkData;
+  const gatherFiles = (result, { files }) => [
+    ...result,
+    ...files.map((file) => `${publicPath}/${file}`),
```
herschel666: Why no `path.join`? Would release you from the duty of trimming slashes.

```diff
- ...files.map((file) => `${publicPath}/${file}`),
+ ...files.map((file) => path.join(publicPath, file)),
```

packages/react/mixins/mixin.browser.js
```diff
@@ -26,14 +26,12 @@ class ReactMixin extends Mixin {
     return createElement(BrowserRouter, props, element);
   }
   render() {
-    const { name } = this.config;
-    const attribute = `data-${name}`;
-    const mountpoint = document.getElementById(name);
-    const isMounted = mountpoint.hasAttribute(attribute);
+    const mountpoint = document.body.firstElementChild;
```
herschel666: This imho not specific enough. Can we be sure 100% of the time, that there's no Google TagManager(-like) script on the page, that inserts elements in a way, that corrupts our way of getting the mountpoint node?

packages/webpack/lib/plugins/stats.js
dmbch: Unfortunately `join` is platform-specific: 
https://nodejs.org/api/path.html#path_path_join_path

We could possibly go for `path.posix.join`, but I always found that being explicit in cases like this had some merit.

packages/react/mixins/mixin.browser.js
dmbch: I agree. What do you propose instead? I would not like to rely on magic strings here - and the previous solution was similarly brittle.

dmbch: How about a `getMountpoint` hook?

packages/webpack/lib/plugins/stats.js
herschel666: Ah, missed the fact that we're moving in a URL context here. So `url-join` was the weapon of choice here. But pulling that in seems a little over the top. So string concatenation would be fine for me.

packages/react/mixins/mixin.browser.js
herschel666: A dedicated hook for that feels a little too much to me. Wouldn't an adequately specific attribue name like `data-application-root` do the trick?

```js
const mountpoint = document.querySelector('[data-application-root]')
```

Besides: we could (maybe should) leave the old ID-attribute for peeps that — for whatever reason — rely on it in their code?!

dmbch: I like the `document.querySelector` approach - will update this PR accordingly.

The previous solution used to rely on `config.name` client-side while allowing people to alter it using `getTemplateData` server-side, making it very *surprising* to use. Besides, I fail to come up with valid use cases for that `id`...

herschel666: > Besides, I fail to come up with valid use cases for that `id`...

Other people might be more creative in that matter. But I guess after all this `id` is one of the internals of Untool & no user should rely on its existence. 🤷‍♂️

dmbch: Done - PTAL, @herschel666

dmbch:
@herschel666 thanks for your review and approval! :rocket:

Do you happen to have an opinion on that `fetchedData` question? I am happy to keep it around for now, but I would love to at least talk it through...

jhiode:
How about creating an issue or PR in hops @dmbch? 🙂

PR #263, opened 2019-02-05
refactor(react): get rid of `fetchedData` prop

dmbch:
Follows up on https://github.com/untool/untool/pull/261
Blocked by https://github.com/xing/hops/pull/789


packages/react/mixins/mixin.server.js
```diff
@@ -44,15 +44,15 @@ class ReactMixin extends Mixin {
       .then(() => this.bootstrap(req, res))
       .then(() => this.enhanceElement(this.element))
       .then((element) =>
-        this.fetchData({}, element).then((fetchedData) => {
+        this.fetchData({}, element).then(() => {
```
jhiode: We could try to flatten the promise chain a little bit:
```js
.then(() => this.bootstrap(req, res))
.then(() => this.enhanceElement(this.element))
.then((element) => this.fetchData({}, element).then(() => element))
.then((element) => {
  const reactMarkup = renderToString(element);
  const fragments = Object.entries(renderStatic()).reduce(
    (result, [key, value]) => ({ ...result, [key]: value.toString() }),
    { reactMarkup, headPrefix: '', headSuffix: '' }
  );
  const assets = getAssets(this.stats, this.context.modules);
  const globals = { _env: this.config._env };
  return { fragments, assets, globals };
})
.then((initialData) => {
```

dmbch: Done.

PR #271, opened 2019-02-15
feat(react): allow import load function in importComponent

robin-drexler:
https://astexplorer.net/#/gist/8aa47f0a6619cf54e99bdf60055bf7d0/b4e7efb45ce31269459f1ae67ed9b50d0e812802

In addition to

```js
importComponent('./foo')
```

also allow 

```js
importComponent(() => import('./foo'))
```

This will allow editors to resolve the imported component. In its current form, for editors it's just a string that they can't follow.

### DIFF
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -53,7 +53,7 @@ export default () => <Status code={404} />;
 
 ### `<Header name value />`
 
-With this component, you can declaratively set arbitrary HTTP headers from your React application.
+With this component, you can declaratively set arbitrary HTTP headers from your React application. On the client side, it is effectively a no-op.
 
 ```javascript
 import { Header } from '@untool/react';
@@ -61,17 +61,29 @@ import { Header } from '@untool/react';
 export default () => <Header name="X-Foo" value="Bar" />;
 ```
 
-### `importComponent(module, [exportName])`
+### `importComponent(module|moduleLoader, [exportName|exportResolver])`
 
 Using the `importComponent` helper, you can asynchronously require components into your application to help you reduce asset sizes. It works similarly to [`react-loadable`](https://github.com/jamiebuilds/react-loadable), but is deeply integrated with `untool`.
 
 ```javascript
 import { importComponent } from '@untool/react';
 
-const Home = importComponent('./home');
+const Home = importComponent('./home', 'Home');
 
 export default () => <Home />;
 ```
+
+Additionally, `importComponent` supports an alternative syntax that helps with editor and type checker integration since it does not rely on plain strings. The snippet below is functionally equivalent to the one above:
+
+```javascript
+import { importComponent } from '@untool/react';
+
+const Home = importComponent(() => import('./home'), ({ Home }) => Home);
+
+export default () => <Home />;
+```
+
+If you do no specify an `exportName` or `exportResolver`, `importComponent` will fall back to the imported modules `default` export.
 
 `importComponent` itself returns a React component supporting some props that enable you to control module loading and (placeholder) rendering.
 
--- a/packages/react/lib/babel.js
+++ b/packages/react/lib/babel.js
@@ -21,17 +21,30 @@ module.exports = ({ types: t }) => ({
         t.assertCallExpression(call);
 
         const argument = call.get('arguments')[0];
-        t.assertStringLiteral(argument);
+
+        let importedComponent;
+        let load;
+
+        if (t.isStringLiteral(argument)) {
+          importedComponent = argument.node;
+          load = t.arrowFunctionExpression(
+            [],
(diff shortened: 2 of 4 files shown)


dmbch:
This looks very promising, thanks! :rocket:

I will have to think this through, though, because I am afraid we are opening ourselves up to at least one edge case: multiple `import()` statements inside that arrow function.

Did you consider using the alternative approach we discussed IRL, e.g.:

```javascript
importComponent(import('./foo'));
```

I must admit that that looks a bit dim semantically, but at least it would eliminate the edge case. We could also alias `importComponent` to make this easier on the eyes...

robin-drexler:
> Did you consider using the alternative approach we discussed IRL, e.g.:

I did, indeed. I decided against it, because imo the transpiled result will be too far away from the input. I.e. users that now about `import()`'s semantics could be confused why the component isn't actually being loaded immediately.

> I will have to think this through, though, because I am afraid we are opening ourselves up to at least one edge case: multiple `import()` statements inside that arrow function. 

True, but you can only return one statement. And the last `import()` will actually be picked up by the babel plugin.

We could, however, think about throwing an error for such cases or even consider creating a hops/untool eslint plugin. Although I'm convinced there should be more than one rule before even start thinking about creating such a plugin.

dmbch:
I could not agree more with regards to a linter plugin: Let us please not do that at this point. Although, thinking even further outside the box, how about building a VSCode/Atom extension to make sure paths passed to `importComponent` are properly identified as such..?

With regards to (assumed) user expectations, I am not quite convinced, personally. Specifically, I am not sure in what way we would be be violating the principle of least surprise here: `import()` is asynchronous after all. Everything we are doing is, from a user’s perspective, an implementation detail, is it not? It *is* being called automatically, after all.

dmbch:
Oh, and by the way: would it not appear somewhat reasonable to pass an arrow function such as this..:

```javascript
() => Promise.all(import('./foo'), import('./bar'))
```

This can of course be dealt with (maybe even in an unbreaking fashion), but I am honestly unsure if the outcome is worth it.

robin-drexler:
> how about building a VSCode/Atom extension to make sure paths passed to `importComponent` are properly identified as such..?

In fact, I already looked into this, but did not pull through. It also felt like re-inventing the wheel because editors are already perfectly capable of resolving modules if one uses the proper mechanics (import/require). 

> It _is_ being called automatically, after all.

It is, but only once the component is supposed to be rendered and not during importing phase as it would be the case if the 

```js
importComponent(import('./foo'));
```

statement was untouched (and like the user sees it).

robin-drexler:
> Oh, and by the way: would it not appear somewhat reasonable to pass an arrow function such as this..

Not for me, no. But that's also something were we could error out should somebody try to something like that.

dmbch:
I am still not convinced, sorry! Nevertheless, I am fine with merging this iff we make sure this is all we additionally support:

```js
importComponent(() => import('./foo'))
```

...and we error out if there is anything else inside the body of that arrow function.

I am just not ready to accept opening us up to a host of edge cases (and addressing them individually) for the sake of slightly improved editor support.

p.s. I *did* notice the implication of `importComponent()` being an improper mechanism ;)

ZauberNerd:
I'm also in favor of this change, but I think we should not spend too much time thinking about how to protect against edge-cases, because if users use that API the wrong way then they can expect it to break - in my opinion we can get away with clarifying this in the docs.

I'd even propose more changes to the API in order to allow type checkers to warn against misuse of the API and therefore would want to have at least the following variants working:

```javascript
const Foo = importComponent(() => import('./foo').then(exports => exports.default));
```

```javascript
const Foo = importComponent(async () => (await import('./foo')).default);
```

```javascript
const Foo = importComponent(() => import('./foo'));

render(<Foo render{({ Component }) => <Component.default />} />);
```

**Edit** But looking at our implementation in depth, it looks like that the only usable case would be the last variant where we overwrite the `render` prop because of how we deal with webpack modules on the server side (https://github.com/untool/untool/blob/master/packages/react/lib/runtime.js#L41-L72)

I'll try to find some time this week to experiment with TypeScript to find out how we could provide types for that case.

dmbch:
@robin-drexler @herschel666 @ZauberNerd how about going with a minimal solution now (only './foo' and `() => import('./foo')` allowed) now and iterate afterwards? That way, we would have something that would give us editor support quite soon.

Afterwards, i.e. when we know what we need to do to get TS support, we can tackle that in another PR.

robin-drexler:
Sounds good to me. If nobody objects I'd implement this early next week. :)

ZauberNerd:
Updated - please take another look @dmbch @robin-drexler @herschel666

PR #274, opened 2019-02-20
use tilde range for inner untool dependencies

jhiode:
In order to make sure that the untool packages that are used in combination does not differ too much I propose to use (at least) tilde ranges. I think it would even be ok to use the exact version as long as we always publish all packages.

IMO what we should use here is the peer dependencies feature, but as that is a breaking change we have to wait for untool 2.0.


dmbch:
I cannot find the relevant PR, but this it the commit where this happened: https://github.com/untool/untool/commit/92551065bd51a2c19cb82843f1c76e031578eae6

From today's perspective, however, I completely agree with you, @jhiode.

dmbch:
I have checked/thought again and did not come up with a satisfactiry solution except for specifying exact dependency versions. Due to our (documented) semver interpretation, I would consider that the best approach.

dmbch:
Oh, I do not necessarily agree with the `peerDependencies` idea, though. Most packages directly require `@untool/core`, meaning that at least that needs to be a proper `dependency`, I think.

@jhiode, would you mind clarifying your reasoning?

ZauberNerd:
I think exact versions would be okay, as long as we keep in mind that we should be careful when updating other dependencies (such as webpack or react).

robin-drexler:
Once concern I have with pinned version is the following scenario:

- our internal hops packages use the external hops in version 11.5.0
- untool gets updated to 1.5.
- hops upgrades untool to 1.5
- hops is released in version 11.5.1
- **our internal packages are not yet updated**
- consumers that use our internal and some external hops packages (like hops-typescript or hops-styled-components can now easily end up with multiple hops and untool versions wreaking havoc

dmbch:
@robin-drexler that could possibly be mitigated by re-exporting the optional external hops packages internally, could it not? I mean we could internally publish a package that essentially pins the versions of any optional external hops packages we might want to provide...

Consumers would then, for example, have to use the preset `@xingternal/hoptional/typescript` instead of `hops-typescript`. And yes, this very approach would imply they have to explicitly configure their presets...

robin-drexler:
> hoptional
🤜 🤛 

> And yes, this very approach would imply they have to explicitly configure their presets...

If we instead published an internal package for each external one, we didn't have to force teams to do that, right?

jhiode:
Regarding `peerDependencies`: With nmp we have to live with the "nested packages" feature. To work around that we have `peerDependencies`. But we have two different kinds of dependencies even in untool. If the package only has some helper functions it is totally fine do define that as dependency. But packages that requires to be shared across multiple other packages we should define them as peer dependency to make that requirement clear instead of relying on the hoisting "optimisation".
So yes, some untool packages can be just dependencies but other should be defined as peer dependencies - but as that is a breaking change we can't do that right now.

So I think tilde range is still the best option we have if we also want to support a mix of internal and external packages.

dmbch:
@robin-drexler true, we can of course mirror the optional externals - it won't be too many, I guess.

@jhiode tilde ranges are unfortunately not supported by Lerna, so if we do not want to try and convince them to add support for them - or build our own tooling - we will probably be out of luck. It will be `exact` or `caret` for now, I am afraid...

Maybe we can at least warn in case of duplicates - using bash, we could add something like this as the `postinstall` hook of our main `@xingternal/hops` package:

```bash
npm ls --parseable 2>/dev/null | awk '!/untool/ && !/hops/' | xargs -L1 sh -c 'basename $1' dummy | sort | uniq -c | grep -v "^ *1 " | sort -rn | awk '{print "Dupe Warning:" $0}'
```

Hat tip: https://stackoverflow.com/questions/35144327/how-to-find-npm-duplicate-packages

jhiode:
Ok, then let's stick with caret range for now. I'm afraid that pinning the versions could make the situation even worse.

dmbch:
@jhiode are you sure? I would have preferred `exact` plus `postinstall`, to be honest.

jhiode:
I'm still planning to provide a `postinstall` hook. Using exact versions could work if developers would use untool directly. But as we have multiple packages that depends on untool I think it might get out of hand very easily. And as the `postinstall` hook would fix the issue anyway I'm not convinced to make our lives any harder with stricter rules 😬

PR #275, opened 2019-02-20
fix(webpack): re-enable node source maps

dmbch:
Currently, node source map support is broken in dev and static builds. This change makes sure it is working again, but it also implies slower (re)builds.


packages/webpack/lib/configs/node.js
```diff
@@ -123,7 +123,7 @@ module.exports = function getConfig(config, name) {
       new EnvironmentPlugin({ NODE_ENV: 'development' }),
     ],
     performance: { hints: false },
-    devtool: isProduction ? 'source-map' : 'cheap-module-eval-source-map',
+    devtool: 'inline-source-map',
```
ZauberNerd: Why can't we use an external source-map?

dmbch: Because we are doing in-memory compilation in static mode and the source map will not be found otherwise.

packages/webpack/lib/shims/server.js
```diff
@@ -1,6 +1,8 @@
 'use strict';
 
-require('source-map-support/register');
+if (!require('module').prototype._compile.__sourceMapSupport) {
```
ZauberNerd: This looks a bit fishy, isn't this handled by the package already? https://github.com/evanw/node-source-map-support/blob/master/source-map-support.js#L17

dmbch: True, that does look bad - and it ought to be handled by the package. I initially thought that the reason might be that we bundle *all the things*, but apparently, it is not. It want(ed) to get this fix shipped, so I went with the solution we have now.

PR #276, opened 2019-02-20
Robustify

dmbch:
I have been planning to add more checks and validation to our public API and to our mixin hooks for a while. In this PR, I have tried a very "mixinable-ish" approach on all our hooks...


packages/core/lib/utils.js
```diff
@@ -5,6 +5,47 @@ const flatten = require('flat');
...
+};
+
+exports.validate = (strategy, ...validators) => {
+  const validate = (data) => validators.forEach((fn) => fn(data));
+  return Object.defineProperty(
+    (functions, ...initialArgs) =>
+      strategy(
+        functions.map((fn) => (...callArgs) => {
+          validate({ before: true, callArgs, initialArgs });
+          const result = fn(...callArgs);
```
herschel666: I'd consistently go down the Promise path here for the sake of readability.

```diff
- const result = fn(...callArgs);
- if (result && typeof result.then === 'function') {
-   return result.then(function(result) {
-     validate({ after: true, result, callArgs, initialArgs });
-     return result;
-   });
- } else {
-   validate({ after: true, result, callArgs, initialArgs });
-  return result;
- }
+ Promise.resolve(fn(...callArgs)).then((result) => {
+   validate({ after: true, result, callArgs, initialArgs });
+   return result;
+ });
```

dmbch: I would certainly like that, but we cannot: we will want to add validation to `core` mixin methods and those are all synchronous.

dmbch:
@herschel666 thanks for your feedback! I came up with what is there now, because I was/am not sure whether we want to have additional validation phases: e.g. we might want to validate `initialArgs` (just once) or `result`...

I also considered adding a validator object like so:

```javascript
const validator = {
  initial({ initialArgs: [element] }) {
    invariant(isValidElement(element), 'Received invalid element');
  },
  call({ callArgs: [element] }) {
    invariant(isValidElement(element), 'Received invalid element');
  },
  intermediate({ result }) {
    invariant(isValidElement(result), 'Returned invalid element');
  },
  result({ result }) {
    invariant(isValidElement(result), 'Returned invalid element');
  },
};
```

At some point, someone might even have experimented with (ab)using `prop-types` in this context, so I am extremely open to API design suggestions here!

ZauberNerd:
Also, I would like to see how this impacts production performance, maybe we should implement this validation strategy for a few "hot" mixin methods and run some load tests to compare the impact they have on production performance.

packages/core/lib/utils.js
```diff
@@ -5,6 +5,47 @@ const flatten = require('flat');
...
 
+const isProduction = process.env.NODE_ENV === 'production';
+
+exports.invariant = (condition, message) => {
+  if (condition) {
+    return;
+  }
+  let error;
+  if (isProduction) {
+    error = new Error('Invariant failed');
```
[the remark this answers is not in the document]
dmbch: Thanks for sharing your opinion, but looking at the numbers, it appears that most projects outside FB use `invariant` similarly to us: `invariant` has 6,7M downloads whereas `babel-plugin-dev-expression` has only 25k. Besides that, that babel plugin does not appear to work in our setup.

Nevertheless, I see no harm in always displaying the message - let us just do that.

packages/react/lib/runtime.js
```diff
@@ -1,79 +1,107 @@
...
+  initialize,
+  internal: { invariant },
+} = require('@untool/core');
 
 exports.render = (element, options) => (...args) => {
   const { render } = initialize({}, element, options);
-  if (!render) {
-    throw new Error("Can't use @untool/react mixin");
-  }
+  invariant(render, "Can't use @untool/react mixin");
```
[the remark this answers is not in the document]
dmbch: Yes, of course we can - and should. Let us agree on an API and then come up with nice and descriptive error messages.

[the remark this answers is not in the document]
dmbch: By the way: At this point, we cannot tell whether the package is installed and just not configured or missing. This message is one of the better thought through ones...

dmbch:
@ZauberNerd with regards to performance: you know how I feel about performance - if we are well above 1kreq/s, we are certainly way more than fine. I mean, [Wikipedia](https://stats.wikimedia.org/EN/TablesPageViewsMonthly.htm) averages at 2,6kreq/s *across all languages*.

That being said, all our hot mixin methods are already being validated in this PR. I just ran a quick, blatantly unscientific `ab` test against my machine and the numbers for `master` and `robustify` were indistinguishable.

packages/react/lib/runtime.js
```diff
@@ -1,79 +1,107 @@
...
+  invariant(isValidElement(element), 'Received invalid element');
```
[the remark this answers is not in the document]
dmbch: I really like the second version. That appears to be a very sensible pattern.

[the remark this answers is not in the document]
dmbch: ...and done.

packages/core/lib/utils.js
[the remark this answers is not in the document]
dmbch: Done!

packages/webpack/mixins/build/mixin.core.js
```diff
@@ -71,17 +74,38 @@ class WebpackBuildMixin extends Mixin {
...
 }
 
 WebpackBuildMixin.strategies = {
-  clean: callable,
-  build: callable,
-  inspectBuild: sequence,
+  clean: validate(
+    callable,
+    ({ length }) => {
+      invariant(length === 0, 'clean(): Received obsolete argument(s)');
```
ZauberNerd: Here again: I'd like this message to be more descriptive.

packages/webpack/mixins/build/mixin.core.js
```diff
@@ -71,17 +74,38 @@ class WebpackBuildMixin extends Mixin {
...
+    },
+    (result, isAsync) => {
+      invariant(isAsync, 'clean(): Did not return a Promise');
+    }
+  ),
+  build: validate(
+    callable,
+    ({ length }) => {
+      invariant(length === 0, 'build(): Received obsolete argument(s)');
```
ZauberNerd: And again: This message is not really descriptive.

packages/webpack/mixins/config/mixin.core.js
```diff
@@ -60,9 +65,39 @@ class WebpackConfigMixin extends Mixin {
 }
 
 WebpackConfigMixin.strategies = {
-  getBuildConfig: callable,
-  collectBuildConfigs: sequence,
-  configureBuild: sequence,
+  getBuildConfig: validate(callable, ([target, baseConfig]) => {
+    invariant(
+      typeof target === 'string',
```
ZauberNerd: Here we could also validate on a concrete value and allow only one-of...

packages/webpack/mixins/config/mixin.core.js
```diff
@@ -60,9 +65,39 @@ class WebpackConfigMixin extends Mixin {
...
+      'getBuildConfig(): Received invalid target string'
+    );
+    invariant(
+      !baseConfig || typeof baseConfig === 'string',
+      'getBuildConfig(): Received invalid baseConfig string'
```
ZauberNerd: In my opinion this message is quite vague too - can we make it more descriptive?

packages/webpack/mixins/config/mixin.core.js
```diff
@@ -60,9 +65,39 @@ class WebpackConfigMixin extends Mixin {
...
+      invariant(
+        isPlainObject(webpackConfig),
+        'configureBuild(): Received invalid webpackConfig object'
+      );
+      invariant(
+        isPlainObject(loaderConfigs),
+        'configureBuild(): Received invalid loaderConfigs object'
+      );
+      invariant(
+        typeof target === 'string',
```
ZauberNerd: And here we could validate on a concrete set of values again, I think?

packages/webpack/mixins/render/mixin.core.js
```diff
@@ -63,7 +66,21 @@ class WebpackRenderMixin extends Mixin {
 }
 
 WebpackRenderMixin.strategies = {
-  getRenderRequests: override,
+  getRenderRequests: validate(
+    override,
+    ({ length }) => {
+      invariant(
+        length === 0,
```
ZauberNerd: Here I'm not sure again, what this message actually means.

packages/webpack/mixins/stats/mixin.core.js
```diff
@@ -43,7 +48,18 @@ class WebpackStatsMixin extends Mixin {
 }
 
 WebpackStatsMixin.strategies = {
-  getBuildStats: callable,
+  getBuildStats: validate(
+    callable,
+    ({ length }) => {
+      invariant(length === 0, 'getBuildStats(): Received obsolete argument(s)');
```
ZauberNerd: Here again we have an unclear message - can we somehow make this more descriptive?

packages/yargs/mixins/main/mixin.core.js
```diff
@@ -15,10 +25,25 @@ class YargsMixin extends Mixin {
 }
 
 YargsMixin.strategies = {
-  registerCommands: sequence,
-  configureCommand: (...args) => sequence(...args) && args[1],
-  handleArguments: sequence,
-  handleError: override,
+  registerCommands: validate(sequence, ([yargs]) => {
+    invariant(
+      yargs && typeof yargs.command === 'function',
```
ZauberNerd: Could we do an `instanceof` check here?

packages/express/mixins/mixin.core.js
```diff
@@ -67,11 +72,62 @@ class ExpressMixin extends Mixin {
...
+    invariant(
+      app && app.handle && app.route,
+      'configureServer(): Received invalid Express app'
+    );
+    invariant(
+      isPlainObject(middlewares) && Object.keys(middlewares).length >= 16,
+      'configureServer(): Received invalid middlewares object'
+    );
+    invariant(
+      typeof mode === 'string',
```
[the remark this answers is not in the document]
dmbch: Nope, that mechanism is meant to be extensible...

packages/express/mixins/mixin.core.js
```diff
@@ -67,11 +72,62 @@ class ExpressMixin extends Mixin {
...
 
 ExpressMixin.strategies = {
-  configureServer: sequence,
-  inspectServer: sequence,
-  runServer: callable,
-  createServer: callable,
-  createRenderer: callable,
+  configureServer: validate(sequence, ([app, middlewares, mode]) => {
+    invariant(
+      app && app.handle && app.route,
```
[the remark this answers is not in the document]
dmbch: `instanceof` what, exactly..? Afaik there is only the `express()` factory, but no exposed constructor. Duck tests are fine, here, I think.

packages/express/mixins/mixin.core.js
```diff
@@ -67,11 +72,62 @@ class ExpressMixin extends Mixin {
...
+      'configureServer(): Received invalid middlewares object'
+    );
+    invariant(
+      typeof mode === 'string',
+      'configureServer(): Received invalid mode string'
+    );
+  }),
+  inspectServer: validate(sequence, ([server]) => {
+    invariant(
+      server && typeof server.listen === 'function',
```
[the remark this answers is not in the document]
dmbch: It might be `http.Server` or `https.Server`, here - again, I consider a duck test sufficient.

packages/express/mixins/mixin.core.js
```diff
@@ -67,11 +72,62 @@ class ExpressMixin extends Mixin {
...
+      'inspectServer(): Received invalid HTTP server instance'
+    );
+  }),
+  runServer: validate(callable, ([mode]) => {
+    invariant(
+      typeof mode === 'string',
```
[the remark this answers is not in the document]
dmbch: No again, this mechanism is also meant to be extensible...

packages/yargs/mixins/main/mixin.core.js
dmbch: `instanceof` what, exactly?

ZauberNerd: Ah sorry, I misread this part..

packages/react/mixins/mixin.browser.js
```diff
@@ -45,10 +50,46 @@ class ReactMixin extends Mixin {
...
+    },
+    (result) => {
+      invariant(
+        isPlainObject(result),
+        'fetchData(): Returned invalid data object'
+      );
+    }
+  ),
+  render: validate(override, ({ length }) => {
+    invariant(length === 0, 'render(): Received unexpected argument(s)');
```
Xiphe: This validation fires in our app (fresh `hops` setup without any custom mixings). When hot module patches are to be applied. `checkArgs` is called with `["./src/app.js"], ["./src/app.js"]` in that scenario.

dmbch: @Xiphe thanks, please see: https://github.com/untool/untool/pull/288

Xiphe: just found it. thanks!

PR #278, opened 2019-02-25
Detect duplicate preset installs

dmbch:
~Since duplicate detection takes quite a lot of time, we will have to decide whether to add it to all presets (as would be technically appropriate) or only to the main `untool` (and `hops`) packages.~

I found [a way](https://github.com/untool/duplitect/blob/9c0808d8e94465e4841d612aaa23f2952a126c85/index.js#L27-L66) to reduce detection time down to a fraction of a second. Thus, I was able to add `duplitect` to all packages. Now we can make sure not `untool` package can be installed more than once into any given project.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -20,6 +20,7 @@
     "cosmiconfig": "^5.0.5",
     "debug": "^4.0.0",
     "dotenv": "^6.0.0",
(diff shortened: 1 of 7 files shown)


dmbch:
I will tackle this topic in a more generalistic PR later.

PR #284, opened 2019-03-05
fix(react): use `require` instead of `import`

herschel666:
…when the `resolveAbsolutePaths`-option is set to `true`.

Also do not include the `moduleId`-property in that case since it is only needed
for the module-ID-related resolving scenario in the context of Webpack.

This fix enables downstream projects to use the result of the Babel plugin
with Node v8. Otherwise this would lead to the syntax error `Unexpected token import`.


herschel666:
Agreed with @dmbch that things are getting a little out of hand here. So closing this PR.

PR #289, opened 2019-03-08
chore: increase test timeouts to stabilize travis

dmbch:


dmbch:
Will have to investigate further.

PR #292, opened 2019-03-08
feat(yargs): improve webpack logging

dmbch:
webpack-logging


packages/webpack/lib/configs/build.js
```diff
@@ -141,8 +141,9 @@ module.exports = function getConfig(config, name) {
       new EnvironmentPlugin({ NODE_ENV: 'development' }),
     ],
     performance: {
```
toomuchdesign: Should @untool provide such "opinionated" defaults?
And also, is there a reason why we configure `performance` field and we keep it turned off with `hints: false`?

packages/yargs/lib/webpack.js
```diff
@@ -1,11 +1,49 @@
 'use strict';
```
toomuchdesign: Thoughts of a freshman: providing unit tests for such a module would extremely help any future maintainer understanding what's going on here. :)

packages/webpack/lib/configs/build.js
dmbch: I think it should - in fact, it always did: Webpack sets both values to 250000, I simply bumped it to 262144 (= 256kB = 0,25MB). I disabled the performance warning output to be able to handle it in `untool` directly: 

https://github.com/untool/untool/blob/007f3e160f5e0b132aec843de6e45dab99c48912/packages/yargs/lib/webpack.js#L17-L28

packages/yargs/lib/webpack.js
dmbch: That is probably true - given the integrative, I/O-heavy nature of most of `untool`, I have postponed the addition of unit tests to a later point. E.g., to comprehensively unit test this plugin thing, we would not just have to mock a logger, but also a major part of webpack's plugin API.

packages/yargs/lib/webpack.js
```diff
@@ -17,32 +55,33 @@ exports.LoggerPlugin = class LoggerPlugin {
...
+
       const hasWarnings = stats.hasWarnings();
       const hasErrors = stats.hasErrors();
-      const duration = prettyMS(stats.endTime - stats.startTime);
+
+      const { assets, errors, warnings, children } = stats.toJson();
+
       if (hasErrors) {
-        this.logger.info(`build target '${name}' failed after ${duration}`);
+        this.logger.info(formatError(name, duration, isRebuild));
```
ZauberNerd: `logger.`**`info`?** for an error?

packages/yargs/lib/webpack.js
```diff
@@ -17,32 +55,33 @@ exports.LoggerPlugin = class LoggerPlugin {
...
       } else {
         if (hasWarnings) {
-          this.logger.info(
-            `build target '${name}' finished with warnings after ${duration}`
-          );
+          this.logger.info(formatWarning(name, duration, assets, isRebuild));
```
ZauberNerd: Same here - **`logger.info` for a warning?

packages/yargs/lib/webpack.js
```diff
@@ -1,11 +1,49 @@
...
+          if (chalk.enabled) {
+            return chalk.red(prettyBytes(size));
+          } else {
+            return `${prettyBytes(size)} !!!`;
+          }
+        } else {
+          return prettyBytes(size);
+        }
+      })();
+      return `${chalk.enabled ? chalk.gray('-') : '-'} ${name} (${prettySize})`;
```
ZauberNerd: Would it make sense to align the columns? Maybe with `\t`? Also because we might have the chunk names in there too in a later iteration.

packages/yargs/lib/webpack.js
```diff
@@ -1,11 +1,49 @@
...
 
 const {
   internal: { BuildError, BuildWarning },
 } = require('@untool/webpack');
 
+const formatAssets = (assets) =>
+  assets
+    .filter(({ name }) => !name.endsWith('.map'))
+    .map(({ name, size, isOverSizeLimit }) => {
+      const prettySize = (() => {
```
ZauberNerd: Why are we using an IIFE here instead of just a nested ternary?
```javascript
const prettySize = isOverSizeLimit
  ? chalk.enabled
    ? chalk.red(prettyBytes(size))
    : `${prettyBytes(size)} !!!`
  : prettyBytes(size);

```

ZauberNerd:
For those interested: This is how the output looks like now:
webpack-logging

Btw, would it make sense to suppress errors that were already reported from another compiler?
As in the example above, we print the same error twice, once for the develop build and another time for the node build.

packages/yargs/lib/webpack.js
herschel666: Also is the ternary needed here? The content (`'-'`) is the same and `chalk` should figure out internally whether to apply the gray color or not.

packages/yargs/lib/webpack.js
dmbch: I consider this an informative message about the result of our compilation - the actual error (or warning) is being dealt with separately:

https://github.com/untool/untool/blob/029fa7242c201b41b8f00e7fdeceed2bdc79395e/packages/yargs/lib/webpack.js#L76-L83

packages/yargs/lib/webpack.js
dmbch: Ditto https://github.com/untool/untool/pull/292#discussion_r264287964

packages/yargs/lib/webpack.js
dmbch: I personally dislike nested ternaries - I generally think they are hard to read, but this instance is rather easy on the eye. Will change it accordingly!

packages/yargs/lib/webpack.js
dmbch: @ZauberNerd I tried using tabs but ultimately gave up as we are dealing with dynamic user provided strings (e.g. project name) here and we cannot know how many tabs we need to align everything. I also considered using something like [`cliui `](https://www.npmjs.com/package/cliui) or [`table`](https://www.npmjs.com/package/table), but ultimately decided to postpone it.

dmbch:
@ZauberNerd I completely see your point with regards to logging the same error multiple times, but on the other hand I would really like to keep error logging stateless, at least for now. Besides that, we are dealing with multiple plugin instances here, so we would have to come up with some sort of global state (inside the module file or in the actual logger implementation).

PR #294, opened 2019-03-10
Improve/stabilize puppeteer implementation

dmbch:


dmbch:
This unfortunately did not work neither...

PR #295, opened 2019-03-10
feat(yargs): tweak express request log output

dmbch:


packages/yargs/lib/request.js
```diff
@@ -0,0 +1,38 @@
...
+morgan.token('pretty-response-time', (req, res) => {
+  if (!req._startAt || !res._startAt) return;
+  const ms =
+    (res._startAt[0] - req._startAt[0]) * 1e3 +
+    (res._startAt[1] - req._startAt[1]) * 1e-6;
+  return prettyMS(Number(ms.toFixed()));
+});
+
+exports.loggerMiddleware = (logger) =>
+  morgan(':method :url :pretty-status :pretty-response-time - :pretty-size', {
```
ZauberNerd: Should we also have a date / time in here?

packages/yargs/lib/request.js
```diff
@@ -0,0 +1,38 @@
...
+const morgan = require('morgan');
+const split = require('split');
+
+const prettyMS = require('pretty-ms');
+const prettyBytes = require('pretty-bytes');
+const chalk = require('chalk');
+
+morgan.token('pretty-size', (req, res) => {
+  const size = res.get('content-length');
+  if (!size) return;
```
ZauberNerd: What does morgan do with `undefined`? Does it just print nothing? Or convert it to a string and actually print `undefined`?

dmbch: It prints a `-` if it gets an `undefined`.

packages/yargs/lib/request.js
dmbch: I do not think so - and both the `tiny` and `dev` default morgan formats have neither.

dmbch: By the, I would only consider this request logging a dev feature, it is not meant for production use. One would probably want to register an additional morgan instance for that.

ZauberNerd: Huh, why don't we use the `dev` preset then? https://github.com/expressjs/morgan#dev

dmbch: Because I wanted to provide properly formatted response sizes..:

https://github.com/untool/untool/blob/f6801345916c386269d11e6c74ae3168eb26fff5/packages/yargs/lib/request.js#L10-L14

...and a single mechanism to manage CLI colors (`chalk` detects Terminal color support and supports `--no-colors` flag IIRC).

packages/yargs/mixins/log/mixin.core.js
```diff
@@ -44,13 +44,8 @@ module.exports = class CLIMixin extends Mixin {
...
     if (mode !== 'static') {
-      const morgan = require('morgan');
-      const split = require('split');
-      app.use(
-        morgan('tiny', {
-          stream: split().on('data', (line) => this.logger.request(line)),
-        })
-      );
+      const { loggerMiddleware } = require('../../lib/request');
+      app.use(loggerMiddleware(this.logger));
```
ZauberNerd: So, you said it's only for development, but won't this include it also for `mode === 'serve'`?

dmbch: Yup. Nothing prevents you from calling `un serve -pv`, but personally, I would prefer another logging solution. In production, I would rather not have app servers log anything request related themselves, t.b.h.

ZauberNerd: So we should guard this with `process.env.NODE_ENV !== 'production'` right?

PR #298, opened 2019-03-11
Modular Logging

dmbch:


packages/express/lib/log.js
```diff
@@ -32,7 +32,7 @@ morgan.token('pretty-response-time', (req, res) => {
   return prettyMS(Number(ms.toFixed()));
 });
 
-exports.loggerMiddleware = (logger) =>
+module.exports = (logger) =>
   morgan(':method :url :pretty-status :pretty-response-time - :pretty-size', {
     stream: split().on('data', (line) => logger.request(line)),
```
ZauberNerd: I would like to not have `logger.request` but instead a more universal method like `logger.verbose` or something which could be used in other contexts too. Because `logger.request` suggests that it can only be used for requests and then people might think to use `logger.request` for their own request logging too.

packages/express/mixins/mixin.core.js
```diff
@@ -12,6 +15,10 @@ const {
 } = require('@untool/core');
 
 class ExpressMixin extends Mixin {
+  constructor(...args) {
+    super(...args);
+    this.enhancedPromise = new EnhancedPromise();
```
ZauberNerd: Could we name this instance variable `serverAddress` or similar instead of `enhancedPromise`, so that we actually know what is in it?

packages/yargs/mixins/log/mixin.core.js
```diff
@@ -30,36 +38,22 @@ module.exports = class CLIMixin extends Mixin {
   handleArguments(argv) {
     this.options = { ...this.options, ...argv };
     const { quiet, verbose } = this.options;
-    this.logger.setLogLevel(logLevels.info + verbose - quiet);
-    this.logger.info(
-      `started in ${process.env.NODE_ENV || 'development'} mode`
-    );
+    const logger = this.getLogger();
+    logger.setLogLevel(logLevels.info + verbose - quiet);
+    logger.info(`started in ${process.env.NODE_ENV || 'development'} mode`);
```
ZauberNerd: This has been irritating me for quite some time - it always says `started in xxx mode`, but maybe it would be nice to change the wording based on which command was executed (`build`, `start`, `serve`).

packages/express/lib/log.js
herschel666: Imho the `info`-level already implies verbosity, so having a `logger.verbose`-method seems redundant to me. But `logger.request` might indeed be a little too specific and thus misleading. What about renaming it to something like `logger.plain`?

dmbch: How about a `logLevel` of `verbose` and a couple of different logger methods like `request` and `hint`? I want to keep the `[<project>:request]` prefix in our output.

packages/express/mixins/mixin.core.js
dmbch: We could call it `serverAddressPromise`, but I decided to keep the naming in line with what we already have: https://github.com/untool/untool/blob/5054511e37a7cb3178f950396d1fde1b8656698a/packages/webpack/mixins/stats/mixin.core.js#L18-L25

packages/express/lib/log.js
herschel666: How about a `logger.custom(prefix, message): void` with a `logLevel` of `verbose`?

ZauberNerd: Other loggers use `debug` or `silly` as method names for more verbose output.

dmbch: I think npm uses these levels:

```json
{ 
  error: 0, 
  warn: 1, 
  info: 2, 
  verbose: 3, 
  debug: 4, 
  silly: 5 
}
```

For `debug` output, we already have an existing mechanism I would like to keep around. And for religious reasons, I refuse to facilitate `silly` actions. ;)

Why don’t we go for these levels: 
```json
{ 
  error: 0, 
  warn: 1, 
  info: 2, 
  verbose: 3
}

File: packages/express/mixins/mixin.core.js [hunk 4]
```diff
@@ -46,6 +56,26 @@ class ExpressMixin extends Mixin {
...
+      }
+    }
+  }
+  inspectServer(server) {
+    if (typeof this.getLogger === 'function') {
+      const logger = this.getLogger();
+      server.on('startup', (address) => {
+        const { basePath = '' } = this.config;
+        const fullAddress = `${address}/${basePath}`;
+        this.serverAddressPromise.resolve(fullAddress);
```
ZauberNerd: Not sure, if we want to resolve to the full URL (with `basePath`) or just `protocol` + `host` + `port`?
/cc @herschel666

packages/express/lib/log.js
dmbch: How about using a `Proxy` for generic log messages?

https://github.com/untool/untool/blob/20dfffdde33467f3fe0d357d31f75cf7a7dab77c/packages/yargs/lib/logger.js#L59-L64

packages/express/mixins/mixin.core.js
dmbch: You are completely right, but I kept the path for logging.

packages/express/mixins/mixin.core.js
```diff
@@ -92,14 +121,30 @@ ExpressMixin.strategies = {
...
-  runServer: validate(callable, ([mode]) => {
+  getServerAddress: validate(
+    callableAsync,
+    ({ length }) => {
+      invariant(
+        length === 0,
+        'getServerAddress(): Received unexpected argument(s)'
+      );
+    },
+    (result, async) => {
```
ZauberNerd: Could we please rename `async` to `isAsync` or similar?

packages/yargs/lib/logger.js
```diff
@@ -47,13 +47,20 @@ module.exports = exports = class Logger {
       console.log(`${prefix} ${message}`);
     }
   }
-  request(message) {
+  verbose(type, message) {
```
ZauberNerd: I'm not too happy with this method having a different signature than the other methods on this class.

dmbch: Understood, will tackle this later in another PR. Specifically, I will replace the `verbose` method with something like this:

```javascript
exports.createLogger = (...args) =>
  new Proxy(new Logger(...args), {
    get(logger, prop) {
      return prop in logger
        ? logger[prop]
        : (message) => {
            const { level, name } = logger;
            const prefix = colorize(`${name}:${prop}`, 'gray');
            if (level >= logLevels.verbose) {
              console.log(`${prefix} ${message}`);
            }
          };
    },
  });
```

Anyhow, since there is `the-doctor` branch based on this one, I will not tackle this right away.

dmbch: This is the actual implementation in #304:

https://github.com/untool/untool/blob/c37129645fdd06e5ab53301c62bca4719890e8c6/packages/info/lib/logger.js#L61-L79

Issue #297, opened 2019-03-11
@babel/polyfill polyfills resolve to wrong `core-js` version on compile time


toomuchdesign:
This issue is a follow-up of https://github.com/xing/hops/pull/819.

@untool/webpack relies on [`@babel/present-env`'s `useBuiltIns` feature](https://babeljs.io/docs/en/babel-preset-env#usebuiltins) which [might break](https://babeljs.io/docs/en/babel-preset-env#usebuiltins-entry) when the globally hoisted `core-js` version doesn't match the one expected by `@babel/polyfill`.

## Solution 1: `core-js` as peer dependency
A solution might consist of declaring `core-js@^2` as `@untool/webpack`'s peer dependency to force the consuming application to keep it as it's own direct dependency

## Solution 2: `core-js v2` dependency check on build
On compile time`@untool/webpack` might ensure that the expected version of `core-js` is available at `node_modules` root and issue a warning message if not.

## Solution 3: custom module resolution
Extend the current custom module resolution to target the polyfill declaration injected by babel and redirect them to the expected `core-js` library.

dmbch:
I think I will go with solution 2 - PR forthcoming.

bxt:
@dmbch How is this supposed to work? A `yarn add core-js@2` solved the problem for me, but I had to bascially Google and click through some GitHub issues to get here. I think there should be some kind of warning?

I just set up a fairly basic Hops project and once I added code splitting I got a lot of errors like this:

```
ERROR in: (webpack)/buildin/global.js
Module not found: Error: Can't resolve 'core-js/modules/es7.symbol.async-iterator' in '/Users/bernhardhaussner/xing/grow-frontend/node_modules/webpack/buildin'
 @ (webpack)/buildin/global.js 1:0-52
 @ ./node_modules/stream-http/index.js
 @ ./node_modules/eventsource/lib/eventsource.js
 @ ./node_modules/@untool/webpack/lib/shims/develop.js
```

Using:

```
$ yarn list --pattern '@untool|core-js'
yarn list v1.10.1
├─ @babel/polyfill@7.4.0
│  └─ core-js@2.6.5
├─ @untool/core@1.5.1
├─ @untool/express@1.5.1
├─ @untool/react@1.5.1
├─ @untool/webpack@1.5.1
├─ @untool/yargs@1.5.1
├─ core-js-compat@3.0.0
│  └─ core-js@3.0.0
├─ core-js-pure@3.0.0
└─ core-js@1.2.7
```

[...]

PR #299, opened 2019-03-11
WIP/RFC: Introduce Doctor/Diagnosis Package

dmbch:
I would like to explore and discuss adding a central package that consolidates all cross-cutting checks (think [`duplitect`](https://github.com/untool/duplitect), #297, config validation etc.) currently scattered across our different packages. CRA appears to follow a similar approach in its [`verify*` scripts](https://github.com/facebook/create-react-app/tree/master/packages/react-scripts/scripts/utils).

Besides that, I would like it to add a `diagnose` command similar to [`brew doctor`](https://docs.brew.sh/Manpage#doctor-options) that runs the aforementioned checks and gathers additional information on the application it is running in (installed presets, env variables, config, file system permissions, package and engine versions etc.).

To be honest, I am still considering if the `untool` package itself might be the best place for all of the above: it already 'knows' about all other packages anyway and is our official entry point. It could, of course, also expose the `doctor` mixin for when it is not being used as a preset.

Another alternative might be to put that doctor stuff back into `@untool/yargs`, but keep up around as a separate mixin.

In any case certain checks such as the one necessitated by #297 will have to be performed in the module context of the respective preset. On the other hand, elaborate mitigation hints ought to be centralized imo to prevent duplication.

### DIFF
--- a/package.json
+++ b/package.json
@@ -137,6 +137,7 @@
         [
           "untool",
           "core",
+          "doctor",
           "express",
           "react",
           "webpack",
--- a/packages/core/lib/config.js
+++ b/packages/core/lib/config.js
@@ -51,11 +51,12 @@ exports.getConfig = ({ untoolNamespace = 'untool', ...overrides } = {}) => {
   const settings = loadConfig(untoolNamespace, pkgData, rootDir);
 
   const raw = merge(defaults, settings, overrides);
-  const { mixins, mixinTypes, configSchema, ...clean } = raw;
+  const { presets, mixins, mixinTypes, configSchema, ...clean } = raw;
   const processed = environmentalize(placeholdify(clean));
 
   const config = {
     ...processed,
+    _presets: presets,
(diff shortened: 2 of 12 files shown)


toomuchdesign:
In terms of **boilerplate** code and and **dependencies** I totally see the value of putting all these checks together in a specific "doctor" bundle.

The **price** to pay is that we'd have to duplicate the knowledge regarding the expected configuration (2 sources of truth: consuming untool package and "doctor" bundle). Which also means that each change to configuration's API has to be reflected in the "doctor" bundle. **Might it have maintenance implications?**

A **middle ground solution** might consist of a "doctor" bundle **providing the diagnose** tools to it's sibling packages and these packages being responsible for telling the "doctor" what they expect. Does it make any sense?

dmbch:
@toomuchdesign my thoughts exactly. Let’s see where this leads us eventually :smile:

dmbch:
Closing for now, will reopen later.

PR #303, opened 2019-03-16
fix(react): switch to supported react router imports

dmbch:
Our fixture breaks with the newest version of React Router (v4.4.0) due to this statement:

https://github.com/untool/untool/blob/1621634bdac34f377af23d11c67c0fbb11d04f31/tests/fixtures/untest/index.js#L3

The relevant change was apparently introduced around React Router [v4.4.0-beta2](https://github.com/ReactTraining/react-router/releases/tag/v4.4.0-beta.2). Additionally, warnings are emitted for our other deep imports from React Router, e.g.:

https://github.com/untool/untool/blob/1621634bdac34f377af23d11c67c0fbb11d04f31/packages/react/lib/runtime.js#L5

This PR addresses these issues

### DIFF
--- a/packages/react/lib/runtime.js
+++ b/packages/react/lib/runtime.js
@@ -2,7 +2,7 @@
 /* global __webpack_modules__, __webpack_require__ */
 
 const { isValidElement, createElement, Component } = require('react');
-const { default: withRouter } = require('react-router-dom/es/withRouter');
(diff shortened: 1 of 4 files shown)


dmbch:
Travis status is red because the canary checks against hops are failing - almost certainly because of the very same changes to React Router we are dealing with in this PR.

PR #304, opened 2019-03-17
The Doctor

### LINKED ISSUE #297: @babel/polyfill polyfills resolve to wrong `core-js` version on compile time

dmbch:
In this PR, I am proposing to introduce a dedicated `@untool/info` package. This package will contain an improved logger (extracted from `@untool/yargs`) and a hook and helpers to run preflight checks.

This branch is based on the `modular-logging` branch as proposed in #298 and closes #297.


packages/info/lib/doctor.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+const define = require('mixinable');
+
+const {
+  sync: { callable: callable, sequence },
```
jhiode: Could be just `{ callable, sequence }`

packages/info/lib/doctor.js
```diff
@@ -0,0 +1,27 @@
...
+} = define;
+
+const strategies = {
+  diagnoseConfig: callable,
+  diagnoseDuplicatePackages: callable,
+  diagnoseInvalidPackages: callable,
```
jhiode: So only the doctor can do the diagnose? (I'm asking because the strategy is callable or a.k.a override).

And do we also want to provide a check to make sure all packages of a specific workspace (like `@untool/*` or `hops-*`) have the same version in common?

packages/info/lib/doctor.js
```diff
@@ -0,0 +1,27 @@
...
+  submitDiagnosis: callable,
+  getDiagnoses: (...args) => {
+    return sequence(...args).then((results) => [].concat(...results));
+  },
+  logDiagnoses: sequence,
```
jhiode: Should we name this `inspectDiagnoses`? Because logging is only one way of handling it, right?

packages/info/mixins/doctor/mixin.core.js
```diff
@@ -0,0 +1,47 @@
...
+  }
+  diagnose({ diagnoseConfig, diagnoseDuplicatePackages }) {
+    diagnoseConfig();
+    diagnoseDuplicatePackages('@untool/*');
+  }
+  bootstrap() {
+    const { doctor } = this;
+    return this.diagnose(doctor).then((results) =>
+      doctor.submitDiagnosis(
+        ...[].concat(...results.filter((result) => result !== undefined))
```
jhiode: Why can't we just write `...results.filter((result) => result !== undefined)`?

packages/info/lib/doctor.js
dmbch: There is no way for anything to actually override strategies/implementations in this context, so yes, it is only this very implementation.

I would very much like to add a check like the one you describe, but maybe after merging this PR?

packages/info/lib/doctor.js
dmbch: I thought about that, too - but, once again, all current diagnoses only log, and I would want to keep it that way.

packages/info/mixins/doctor/mixin.core.js
dmbch: Because we must expect the individual results to arrays, so we would get an array of arrays if we did what you propose. We need to flatten that.

packages/info/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Installation
+
+```bash
+$ yarn add @untool/info # OR npm install @untool/info
+```
+
+## CLI
+
+`@untool/info` does not define any commands of its own, but only adds some global CLI flags to control console output. Using the flags `-v`/`--verbose` and `-q`/`--quiet` once or multiple times, you can de- and increase the log output. Default log level is `info`, passing a single `-q` will reduce it to `warning` while a single `-v` will bump it to `verbose`. Passing `-qqq` will silence all output completely.
```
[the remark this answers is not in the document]
dmbch: I *think* it is... I know the `-qq` from [apt-get](https://linux.die.net/man/8/apt-get) and the `-vv` (and even `-vvv`) from [tcpdump](https://linux.die.net/man/8/tcpdump)...

packages/info/README.md
```diff
@@ -0,0 +1,76 @@
+# `@untool/info`
+
+[![travis](https://img.shields.io/travis/untool/untool/master.svg)](https://travis-ci.org/untool/untool)&nbsp;[![npm](https://img.shields.io/npm/v/@untool%2Finfo.svg)](https://www.npmjs.com/package/@untool%2Finfo)
```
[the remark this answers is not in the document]
dmbch: Absolutely true! Nice catch - I now fixed it in almost all our READMEs.

PR #307, opened 2019-03-19
refactor(info): remove obsolete getDiagnoses hook

dmbch:


jhiode:
Why is it obsolete? My understanding was that you could use this method to e.g. collect all results in order to do something else then just logging them.

dmbch:
Theoretically yes, practically no: it is just defined on the `doctor` which is a private mixing property of the `DoctorMixin`.

PR #312, opened 2019-03-22
 feat(webpack): add hot reloading on win32

### LINKED ISSUE #311: Webpack: Signal based HMR crashes on windows
While initially running my app compiles just fine, however on editing any of the app's files, the watcher crashes:
```
untest:error TypeError [ERR_UNKNOWN_SIGNAL]: Unknown signal: SIGUSR2

at process.kill (internal/process/per_thread.js:193:15)

at Watching.compiler.watch [as handler] (.\node_modules\@untool\webpack\lib\middlewares\render.js:50:19)

at compiler.hooks.done.callAsync (.\node_modules\webpack\lib\Watching.js:98:9)

at AsyncSeriesHook.eval [as callAsync] (eval at create (.\node_modules\tapable\lib\HookCodeFactory.js:32:10), <anonymous>:15:1)

at Watching._done (.\node_modules\webpack\lib\Watching.js:97:28)

at compiler.emitRecords.err (.\node_modules\webpack\lib\Watching.js:72:19)

at Compiler.emitRecords (.\node_modules\webpack\lib\Compiler.js:449:39)

at compiler.emitAssets.err (.\node_modules\webpack\lib\Watching.js:53:20)

at hooks.afterEmit.callAsync.err (.\node_modules\webpack\lib\Compiler.js:435:14)

at AsyncSeriesHook.eval [as callAsync] (eval at create (.\node_modules\tapable\lib\HookCodeFactory.js:32:10), <anonymous>:15:1)
```

The signal `SIGUSR2` is not defined in Windows environments. I'm not sure if there is a way to kill the process on windows by another call. Seems like we need to fallback to `webpack/hot/poll` on windows, as suggested here: https://github.com/sysgears/apollo-universal-starter-kit/issues/14

dmbch:
This is more of an RFC at this point - we will certainly have to discuss implications. I did not, however, manage to get HMR working using `webpack/hot/poll` - and `webpack/hot/signal`, i.e. what we currently use, is not supported on win32.

Closes #311


dmbch:
Since it appears to be possible to override the signal `webpack/hot/signal` is listening for...

https://github.com/webpack/webpack/blob/b41e2c68a5d707bd2d315b7b10f6b714c78f0474/hot/signal.js#L48

...and since we might (ab)use one of the [signals supported on win32](
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/signal?view=vs-2017#remarks), we might be able to fix this in a less problematic manner.

dmbch:
Will be fixed in #314

PR #314, opened 2019-03-25
Fix hot module reloading for Windows / Edge and Node 11.11.0

### LINKED ISSUE #311: Webpack: Signal based HMR crashes on windows

ZauberNerd:
This PR contains two parts - first it switches to using custom events instead of signals to make HMR work for the Node build, then it includes an EventSource polyfill to make HMR work in the browser.

Closes #311


packages/webpack/lib/configs/develop.js
```diff
@@ -83,6 +83,7 @@ module.exports = function getConfig(config, name) {
     mode: 'development',
     context: config.rootDir,
     entry: [
+      require.resolve('../shims/event-source'),
```
dmbch: We can add the actual `eventsource` package here right away instead of adding another shim: it is being bundled anyway and will only attach itself to `window` if missing:

https://github.com/EventSource/eventsource/blob/82d38b0b0028ba92e861240eb6a943cbcafc8dff/lib/eventsource-polyfill.js#L5

PR #315, opened 2019-03-25
refactor(info): get rid of word wrapping

dmbch:


packages/info/lib/logger.js
```diff
@@ -1,19 +1,9 @@
 'use strict';
 /* eslint-disable no-console */
 
-const { EOL } = require('os');
-
 const chalk = require('chalk');
-const wordWrap = require('word-wrap');
```
ZauberNerd: I don't see an updated package.json in this PR - shouldn't we remove this dependency?

dmbch:
I cannot technically approve this PR, not even @ZauberNerd’s change - but essentially I do. :shipit:

PR #316, opened 2019-03-26
[RFC]: introduce "absolutePath" keyword for config schema validation

ZauberNerd:
This PR creates a [custom ajv keyword](https://github.com/epoberezkin/ajv/blob/master/CUSTOM.md)
called `absolutePath` which can be set to `true` or `false` depending on
whether the value must or must not be an absolute path.


dmbch:
@jhiode URI support is built-in: https://json-schema.org/understanding-json-schema/reference/string.html#resource-identifiers

PR #317, opened 2019-03-27
[RFC]: Add chunk names to the file names of dynamically imported components

ZauberNerd:
chunk-names


packages/react/lib/babel.js
```diff
@@ -1,5 +1,13 @@
 'use strict';
 
+const path = require('path');
+
+function getName(name, filename, rootDir) {
+  return path
+    .relative(rootDir, path.join(path.dirname(filename), name))
+    .replace(/([^\w])/g, (match) => (match === '/' ? '~' : ''));
```
herschel666: I'd personally keep dashes: `/([^\w-])/g`.

ZauberNerd:
Not relevant right now.

PR #319, opened 2019-03-29
Add missing exports to dist package

ZauberNerd:


packages/react/index.js
```diff
@@ -3,6 +3,8 @@
 exports.internal = {
   babelPlugin: require('./lib/babel'),
   template: require('./lib/template'),
+  runtime: require('./lib/runtime'),
 };
+exports.internal.babelPlugin.path = require.resolve('./lib/babel');
```
dmbch: If we add some property to `babelPlugin`, I would prefer it happening in `./lib/babel`.

packages/webpack/index.js
```diff
@@ -8,8 +8,9 @@ const createStatsMiddleware = require('./lib/middlewares/stats');
 const { RenderPlugin } = require('./lib/plugins/render');
 const { StatsPlugin, StatsFilePlugin } = require('./lib/plugins/stats');
 
-const { BuildError, BuildWarning } = require('./lib/utils/errors');
+const { BuildError } = require('./lib/utils/errors');
 const configLoader = require('./lib/utils/loader');
+configLoader.path = require.resolve('./lib/utils/loader');
```
dmbch: Same here: If we add some property to `configLoader`, I would prefer it happening in `./lib/utils/loader`.

PR #320, opened 2019-03-29
ci: run tests serially stabilize them

dmbch:


dmbch:
Ok, this obviously did not work..: https://travis-ci.org/untool/untool/builds/513139059

PR #322, opened 2019-04-01
feat(info): make logger methods behave like console w.r.t. args

dmbch:


tests/snapshots/build-p.js.md
```diff
@@ -771,7 +771,7 @@ Generated by [AVA](https://ava.li).
           logger: {
             level: 0,
             name: 'untest',
-            workspace: '.',
+            pathRegExp: /\/Users\/daniel\.dembach\/src\/untool\/untool/g,
```
ZauberNerd: This should not contain absolute paths because it will break on other machines..

packages/info/lib/logger.js
```diff
@@ -13,43 +15,44 @@ const logLevels = { error: 0, warn: 1, info: 2, verbose: 3 };
...
-  error(error) {
-    const { level, name, workspace } = this;
+  error(error, ...args) {
+    const { level, name, getCleanMessage } = this;
     const prefix = colorize(`${name}:error`, 'red');
-    const message = String(error.stack || error).replace(
-      new RegExp(escapeRegExp(workspace), 'g'),
-      '.'
-    );
+    const message = getCleanMessage(error);
```
ZauberNerd: Shouldn't we do this for all values in `...args` because any of them could be an error?
We could iterate through the arguments and check for `instanceof Error` or something and then use the `this.getCleanMessage()` method to format the error accordingly?

dmbch: @ZauberNerd I think we decided IRL not to tackle this a this point since we are already mimicking `format`/`inspect` in the current implementation.

packages/webpack/lib/utils/errors.js
```diff
@@ -2,10 +2,13 @@
...
 
-class BuildError {
+const stripAnsi = require('strip-ansi');
+
+class BuildError extends Error {
   constructor(stack) {
+    super();
     this.name = this.constructor.name;
-    this.stack = `${this.name} in ${stack.replace(/\033?\[[0-9]{1,2}m/g, '')}`;
+    this.stack = `${this.name}: ${stripAnsi(stack)}`;
```
ZauberNerd: :heart: :+1: :rocket:

PR #324, opened 2019-04-08
Clean up doctor hook names

ZauberNerd:
This is more of a personal preference, but in my opinion punny names
should not be part of actual source code / APIs.

Also: Before it was unclear from the hook names alone, what they
actually do. I think now the names make a bit more sense and the reader
can infer from the name, what they will do.


packages/info/lib/diagnoses/config.js
```diff
@@ -2,23 +2,23 @@
...
+    const { _warnings } = this.config;
+    this.warnings = _warnings;
   }
-  logDiagnoses(logger) {
-    const { diagnoses } = this;
-    if (diagnoses.length) {
-      const warnings = diagnoses.map(
-        (diagnosis) => `${chalk.yellow('-')} ${diagnosis}`
-      );
+  logMessages(logger) {
```
toomuchdesign: I would expect `logWarnings` but we used this syntax for all the rest.

packages/info/lib/diagnoses/duplicates.js
```diff
@@ -3,25 +3,24 @@
 const chalk = require('chalk');
 const detectDuplicates = require('duplitect');
 
-module.exports = class DuplicatePackagesDiagnosisMixins {
+module.exports = class DuplicatePackagesMixins {
```
toomuchdesign: Since it's a naming-related PR I'll go nitpicky 😄. `DuplicatePackagesMixins` sounds like we're duplicating stuff while here we want to **detect** when something is duplicated. Shall we use somehow `detect`?

packages/info/lib/diagnoses/config.js
ZauberNerd: It is not limited to warnings, it could be errors, warnings, infos or something else, therefore I named it `logMessages`

packages/info/lib/diagnoses/duplicates.js
ZauberNerd: done

packages/info/lib/diagnoses/config.js
dmbch: How about `logResults`?

packages/info/lib/diagnoses/generic.js
```diff
@@ -1,29 +1,29 @@
...
       if (handler) {
-        diagnoses.unshift(handler);
+        messages.unshift(handler);
       }
       handler = defaultHandler;
     }
-    if (diagnosesMap.has(handler)) {
-      diagnosesMap.set(handler, [...diagnosesMap.get(handler), ...diagnoses]);
+    if (messagesMap.has(handler)) {
+      messagesMap.set(handler, [...messagesMap.get(handler), ...messages]);
```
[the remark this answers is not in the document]
dmbch: @herschel666 What is it that you find so very awkward about this? Would this be at least slightly better?

```javascript
if (messagesMap.has(handler)) {
  messagesMap.get(handler).push(...messages);
} else {
  messagesMap.set(handler, [...messages]);
}
```

packages/webpack/mixins/build/mixin.core.js
```diff
@@ -81,13 +83,41 @@ class WebpackBuildMixin extends Mixin {
...
     );
   }
-  diagnose({ diagnoseDuplicatePackages, diagnoseInvalidPackages }) {
-    diagnoseDuplicatePackages('webpack');
-    diagnoseInvalidPackages(__dirname, '@babel/polyfill');
-    diagnoseInvalidPackages(
-      dirname(require.resolve('@babel/polyfill/package.json')),
-      'core-js'
-    );
+  checkHoistedVersion(context, packageName) {
```
dmbch: Let us please not introduce non-hook methods to mixins - see [docs](https://github.com/untool/untool/blob/master/packages/core/README.md#api):

> While it is technically possible to define non-mixin utility methods on your mixin, doing so is strongly discouraged.

packages/webpack/mixins/build/mixin.core.js
```diff
@@ -81,13 +83,41 @@ class WebpackBuildMixin extends Mixin {
...
+        dirname(require.resolve('@babel/polyfill/package.json')),
+        'core-js'
+      ),
+    ]
+      .filter(Boolean)
+      .map((message) => `${chalk.yellow('-')} ${message}`);
+
+    const logger = this.getLogger();
+    if (warnings.length) {
+      collectResults(
```
dmbch: When designing that API, this is how I intended it to be used..:

```javascript
collectResults(
  (warnings, { warn, hint }) => {
    if (warnings.length) {
      warn('Invalid package version(s) detected:\n' + warnings
        .map((message) => `${chalk.yellow('-')} ${message}`)
        .join('\n'));
      hint(
        'Fix invalid package version(s):\n' +
          'Please use your package manager of choice (npm or yarn) to ' +
          'install the package versions mentioned above directly into your ' +
          'project. Not doing so will almost certainly lead to hard to ' +
          'debug issues with your application.'
      );
    }
  },
  warnings
);
```

This way, warnings and hints of concerning a specific type of problem will be grouped.

dmbch: Oh, and I forgot to mention that that handler function is rather self-contained, making it a candidate for extraction into a reusable, separate helper function.

packages/info/lib/diagnoses/config.js
ZauberNerd: I've gone with `logResults`

PR #326, opened 2019-04-09
Normalize path handling

ZauberNerd:


packages/react/mixins/mixin.browser.js
```diff
@@ -24,9 +25,12 @@ class ReactMixin extends Mixin {
     this.element = element;
   }
   enhanceElement(element) {
+    const basename = ensureLeadingSlash(
+      trimTrailingSlash(this.config.basePath)
+    );
     const props = {
       ...this.options.router,
-      basename: this.config.basePath,
+      basename: basename === '/' ? '' : basename,
```
dmbch: Would switching the `pathifist` helper application order not allow us to remove this extra step..?

```
> trimTrailingSlash(ensureLeadingSlash('/foo/'))
'/foo'
> trimTrailingSlash(ensureLeadingSlash('/foo'))
'/foo'
> trimTrailingSlash(ensureLeadingSlash('foo/'))
'/foo'
> trimTrailingSlash(ensureLeadingSlash('/'))
''
```

packages/react/mixins/mixin.server.js
```diff
@@ -36,10 +37,13 @@ class ReactMixin extends Mixin {
...
   enhanceElement(element) {
     const { pathname, search } = this.url;
+    const basename = ensureLeadingSlash(
+      trimTrailingSlash(this.config.basePath)
+    );
     const props = {
       ...this.options.router,
       location: { pathname, search },
-      basename: this.config.basePath,
+      basename: basename === '/' ? '' : basename,
```
dmbch: Same here.

PR #333, opened 2019-04-16
[WIP]: fork webpack compiler in render middleware

ZauberNerd:
By forking the render middleware compiler into a separate process we can make use of multi-core CPUs to speed up compilation (especially helpful for incremental rebuilds).


tests/snapshots/build-p.js.md
```diff
@@ -63,6 +63,9 @@ Generated by [AVA](https://ava.li).
             'string',
           ],
         },
+        _overrides: {
```
dmbch: I would prefer it if we moved the snapshot updates to a separate commit for this whole branch.

packages/webpack/lib/utils/compiler.js
```diff
@@ -0,0 +1,53 @@
...
+    });
+  } else {
+    const { path, filename } = config.output;
+    callback({
+      name: 'success',
+      data: join(path, filename),
+    });
+  }
+}
+module.exports.compilerHandler = compilerHandler;
```
dmbch: Why do we need to export this? If we do not, I would prefer to keep it private.

packages/webpack/lib/utils/compiler.js
```diff
@@ -0,0 +1,53 @@
...
+  const compiler = webpack(webpackConfig);
+  compiler.outputFileSystem = new MemoryFS();
+  sourceMapSupport.install({ environment: 'node', hookRequire: true });
+  compiler.run(
+    compilerHandler.bind(null, webpackConfig, (message) => {
+      callback(message, compiler);
+    })
+  );
+}
+module.exports.createCompiler = createCompiler;
```
dmbch: `exports.createCompiler` is sufficient here.

packages/webpack/lib/utils/compiler.js
```diff
@@ -0,0 +1,53 @@
...
+  const compiler = webpack(webpackConfig);
+  compiler.hooks.watchRun.tap('RenderMiddleware', () =>
+    callback({ name: 'watchRun' })
+  );
+  compiler.watch(
+    webpackConfig.watchOptions,
+    compilerHandler.bind(null, webpackConfig, callback)
+  );
+}
+module.exports.createWatchCompiler = createWatchCompiler;
```
dmbch: `exports.createWatchCompiler` is sufficient here.

packages/webpack/lib/utils/compiler.js
```diff
@@ -0,0 +1,53 @@
+'use strict';
+
+const { join } = require('path');
+const webpack = require('webpack');
+const MemoryFS = require('memory-fs');
+const sourceMapSupport = require('source-map-support');
+
+function compilerHandler(config, callback, compileError, stats) {
```
dmbch: I am really not too keen on this API, sorry! If we had to keep things callback-style, why not stick to the tried-and-tested Node-style callbacks (`(err, res) => {}`)?

Generally, I would prefer this thing to directly returning an `EnhancedPromise`, if at all possible, instead of relying on custom data structures and magic strings.

tests/develop.js
```diff
@@ -38,7 +38,7 @@ test('server lifecycle hooks', (t) =>
     ])
   ));
 
-test('browser lifecycle hooks', (t) =>
+test.serial('browser lifecycle hooks', (t) =>
```
dmbch: Please be so kind as to move test updates to a separate commit.

tests/snapshots/build.js.md
```diff
@@ -1417,9 +1417,6 @@ Generated by [AVA](https://ava.li).
         maxEntrypointSize: 5242880,
       },
       plugins: [
```
dmbch: Please move snapshot updates into a different commit.

dmbch:
@ZauberNerd I think I was too involved in this to provide a proper review, sorry!

PR #343, opened 2019-05-02
feat(react): introduce renderToFragments mixin hook

jhiode:
In order to be able to override and re-use the `renderTostring` and `renderStatic` functions in hops-graphql we've introduced a new mixin hook. We've also improved some code on the way.

We need this hook because we need to combine the `getDataFromTree` with the actual `renderToString` function in order to have the error handling of Apollo errors in the same render call that is used for the SSR content.


packages/react/mixins/mixin.server.js
```diff
@@ -59,13 +59,16 @@ class ReactMixin extends Mixin {
...
+    return renderToFragments(element);
+  }
   render(req, res, next) {
     Promise.resolve()
       .then(() => this.bootstrap(req, res))
       .then(() => this.enhanceElement(this.element))
       .then((element) => this.fetchData({}, element).then(() => element))
-      .then((element) => {
-        const fragments = renderToFragments(element);
+      .then((element) => this.renderToFragments(element))
```
dmbch: Why not..:

```javascript
.then((element) =>
  this.fetchData({}, element).then(() => this.renderToFragments(element))
)
```

packages/react/mixins/mixin.server.js
```diff
@@ -9,7 +9,7 @@ const isPlainObject = require('is-plain-object');
 
 const {
   override,
```
dmbch: Please call this `overrideSync` to make the distinction clear - see also..:

https://github.com/untool/untool/blob/382de1056c68a16d0fe03db642ed65d4d0574b53/packages/express/mixins/mixin.core.js#L7-L10

PR #352, opened 2019-05-09
Include StatsFilePlugin for all one-off builds

ZauberNerd:
All commands that trigger a one-off build, such as:

  - `un build`
  - `un build -p`
  - `un build -s`
  - `un build -ps`
  - `un start -p`
  - `un start -ps` Should also include the StatsFilePlugin for all node

builds in order to be able to write asset information to the disk.


packages/webpack/mixins/stats/mixin.core.js
```diff
@@ -29,14 +29,9 @@ class WebpackStatsMixin extends Mixin {
...
-      this.options.static &&
-      (process.env.NODE_ENV === 'production' ||
-        this.options._.includes('build'));
-    if (
-      target === 'node' &&
-      (isStaticBuild || process.env.NODE_ENV === 'production')
-    ) {
+    const isBuild =
+      this.options._.includes('build') || process.env.NODE_ENV === 'production';
+    if (target === 'node' && isBuild) {
```
ZauberNerd: This is supposed to catch all conditions that are the opposite of: `mode === 'develop'` as is used here:
- https://github.com/untool/untool/blob/master/packages/webpack/mixins/render/mixin.core.js#L38
- https://github.com/untool/untool/blob/master/packages/webpack/mixins/render/mixin.core.js#L41

PR #355, opened 2019-05-09
Improve log mixin / child process handling

ZauberNerd:


packages/info/mixins/log/mixin.core.js
```diff
@@ -15,6 +15,7 @@ class LogMixin extends Mixin {
   constructor(...args) {
     super(...args);
     this.logger = createLogger(this.config);
+    this.handleArguments({});
```
dmbch: Why not `this.handleArguments(this.options)`?

packages/info/mixins/log/mixin.core.js
```diff
@@ -39,11 +40,11 @@ class LogMixin extends Mixin {
   }
   handleArguments(argv) {
     this.options = { ...this.options, ...argv };
-    const { quiet, verbose, _: commands } = this.options;
+    const { quiet = 0, verbose = 0, _: commands = [] } = this.options;
```
dmbch: We could also move this up to the constructor.

ZauberNerd: Can you explain a bit more what you mean by that?

dmbch: It would not necessarily make things more readable, but it would probably make that additional `handleArguments`-call obsolete (at least if we also added similar code to all other mixins that are interested in `options`).

ZauberNerd: I'm sorry, I don't understand. Can you maybe write a small example of what you would want to achieve?
As far as I can tell, we need the `handleArguments` call in the constructor, because it is responsible for setting the `logLevel` and we want to be able to set the logLevel in two cases - executing `un ..` should set the log level, that's why we have the mixin hook, that will be called through yargs and we wan to be able to set the log level through `initialize(config, options)` (or `configure(config, options)` on some public API), that's why we need to call the `handleArguments` method in the constructor.

dmbch: I agree, you are right here!

packages/info/mixins/log/mixin.core.js
```diff
@@ -39,11 +40,11 @@ class LogMixin extends Mixin {
...
     const command = commands.join(' ');
     const logger = this.getLogger();
     logger.setLogLevel(logLevels.info + verbose - quiet);
-    if (typeof process.send === 'undefined') {
+    if (command) {
```
dmbch: 💚

packages/webpack/lib/utils/compiler.js
```diff
@@ -77,9 +77,10 @@ exports.createWatchCompiler = createWatchCompiler;
 process.on('message', (message) => {
   if (message.name !== 'start') return;
   const { buildConfigArgs, overrides, options } = message;
-  const { handleArguments, getBuildConfig } = initialize(overrides);
-  handleArguments(options);
-  const webpackConfig = getBuildConfig(...buildConfigArgs);
+  const webpackConfig = configure(overrides, {
+    ...options,
+    _: [],
```
dmbch: Is this not obsolete as we have the [default](https://github.com/untool/untool/pull/355/files#diff-a4275fbdb28414118289ba0f39b70d04R43)?

ZauberNerd: Nope, we explicitly want to set it to an empty array to prevent the log mixin to log "running '...' in '...' mode". Because the options that we get passed in here, do contain the command of the parent process.

PR #362, opened 2019-05-13
fix(info): hide starting message if no command is set

ZauberNerd:


ZauberNerd:
@dmbch travis is green - just the status report has not arrived at GitHub.
Please merge and release :)

PR #363, opened 2019-05-15
Reenable asset size limits

ZauberNerd:


packages/webpack/lib/plugins/log.js
```diff
@@ -69,6 +70,15 @@ exports.LoggerPlugin = class LoggerPlugin {
           this.logger.info(formatWarning(name, duration, assets, isRebuild));
         } else {
           this.logger.info(formatSuccess(name, duration, assets, isRebuild));
+          const assetsOverSizeLimit = assets.some(
+            ({ isOverSizeLimit }) => isOverSizeLimit
+          );
+          if (assetsOverSizeLimit) {
+            this.logger.warn('Some assets exceed the recommended size limit!');
+            this.logger.hint(
+              'You can limit the size of your bundles by using our importComponent() helper to lazy load some parts of your application.'
```
ZauberNerd: I would have liked to put a link to the documentation of `importComponent` here, but the problem is, that we don't know whether to link to the Hops docs or untool docs, because the code examples all show different packages from which to import the `importComponent` function.
For untool: `import { importComponent } from 'untool';`
For Hops: `import { importComponent } from 'hops';`

ZauberNerd:
Please take a look @herschel666 @dmbch

PR #364, opened 2019-05-16
feat(webpack): print explanatory message if bundles exceed size limits

ZauberNerd:


packages/webpack/lib/plugins/log.js
```diff
@@ -69,6 +69,15 @@ exports.LoggerPlugin = class LoggerPlugin {
           this.logger.info(formatWarning(name, duration, assets, isRebuild));
         } else {
           this.logger.info(formatSuccess(name, duration, assets, isRebuild));
+          const assetsOverSizeLimit = assets.some(
+            ({ isOverSizeLimit }) => isOverSizeLimit
+          );
+          if (assetsOverSizeLimit) {
+            this.logger.warn('Some assets exceed the recommended size limit!');
+            this.logger.hint(
+              'You can limit the size of your bundles by using our importComponent() helper to lazy load some parts of your application.'
```
jhiode: I would like to omit the usage of `yours` and `ours`.

Alternative: "You can limit the bundle size by using the importComponent() helper..."

packages/webpack/lib/plugins/log.js
```diff
@@ -70,6 +70,15 @@ exports.LoggerPlugin = class LoggerPlugin {
           this.logger.info(formatWarning(name, duration, assets, isRebuild));
         } else {
           this.logger.info(formatSuccess(name, duration, assets, isRebuild));
+          const assetsOverSizeLimit = assets.some(
+            ({ isOverSizeLimit }) => isOverSizeLimit
+          );
+          if (assetsOverSizeLimit) {
+            this.logger.warn('Some assets exceed the recommended size limit!');
+            this.logger.hint(
+              'You can limit the bundle size by using the importComponent() helper to lazy load some parts of the application.'
```
dmbch: I am not too keen on this hint as it is as there are plenty of other options to limit bundle sizes (`import()`/`require.ensure()`, dependency deduplication or elimination, etc).

This wording makes it sound like `importComponent` was the only way to limit bundle size. If we are convinced we need a hint like this, how about `Please consider using import() or the importComponent() helper to lazy load some parts of the application`?

packages/webpack/lib/plugins/log.js
```diff
@@ -70,6 +70,15 @@ exports.LoggerPlugin = class LoggerPlugin {
           this.logger.info(formatWarning(name, duration, assets, isRebuild));
         } else {
           this.logger.info(formatSuccess(name, duration, assets, isRebuild));
+          const assetsOverSizeLimit = assets.some(
```
jhiode: I think we should print this message even if we have other warnings...
And without the `-v` flag the hint is not visible, but I guess that is intentional.

dmbch: @jhiode we are considering raising the log level in non-production environments. That would make these hints visible in development mode.

packages/webpack/lib/plugins/log.js
ZauberNerd: I changed to wording to your suggestion for now. Maybe we can find a way to have a readme section / wiki article on more in-depth knowledge for analysing and reducing the bundle sizes later on.

packages/webpack/lib/plugins/log.js
ZauberNerd: @jhiode I moved the warning / hint one level up, so it will be printed even if the build has other warnings.

PR #365, opened 2019-05-16
refactor(info): increase default log level in development mode

ZauberNerd:
log-output

This also shows all request logs, not sure, if that is something that we actually want?
Maybe we need another, lower level, such as `silly` ;)
Or we could implement `hint` with a log level of `info`, so that they will always be visible.


ZauberNerd:
I think these hints provide benefits - why else would we add them?
If we don't want to introduce a new loglevel, I would propose that the doctor checks whether warnings were generated and if so, it should print another line, saying:
`This command produced $n warnings. Execute the command again with "-v" to see additional hints on how to resolve those.`.

dmbch:
Of course the hints are somewhat useful - I am just opposed to printing them by default: one will need to see the warnings themselves every time they are triggered, but the hints only really provide benefits the first time they are generated. After having gotten a hint once, a user will already know what to do about a given type of warning.

Besides that, I would personally prefer improving our docs as opposed to adding a meta-hint such as the one you are proposing. Anyway, feel free to open another PR to deal with this the way you see fit!

PR #366, opened 2019-05-17
feat(core): allow to override rootDir

jhiode:
In order to create (core) mixins in unit tests it would be helpful to change the directory in which untool tries to find the app.


dmbch:
Closing this for now as discussed above and irl. Let’s revisit if we come up with a comprehensive approach!

PR #380, opened 2019-06-12
chore: remove version pinning workaround for canarist hops

ZauberNerd:


ZauberNerd:
Whoops, I mixed up my commits - this is the one, that will make this workaround unnecessary: https://github.com/xing/canarist/pull/8, so after we upgrade canarist, we should be able to merge this pr.

PR #381, opened 2019-06-12
chore: tweak renovate range strategies

dmbch:


package.json
```diff
@@ -102,16 +102,29 @@
...
-    "ignoreDeps": [
-      "prettier"
-    ],
     "lockFileMaintenance": {
       "enabled": true
     },
-    "pinVersions": false,
-    "prCreation": "not-pending",
+    "ignoreDeps": [
+      "prettier"
```
ZauberNerd: Huh, why are you ignoring prettier updates?

package.json
```diff
@@ -102,16 +102,29 @@
...
+    ],
     "semanticCommits": true,
-    "semanticPrefix": "chore:"
+    "semanticCommitScope": "{{parentDir}}",
```
ZauberNerd: I suppose this will not work, as dependencies can be updated in multiple packages and you don't want to have separate commits / PRs, because then your dependencies will run out of sync, which might cause trouble for some dependencies.

package.json
dmbch: I always did, to be honest, as prettier has a rather interesting versioning strategy..:

> We recommend pinning an exact version of prettier in your package.json as we introduce stylistic changes in patch releases.

This is taken straight from the prettier [docs](https://prettier.io/docs/en/install.html).

package.json
dmbch: As this appears to be a rather straightforward/standard use case, I would expect renovate to be able to deal with this. If it turns out to be incapable of this, I would not mind getting rid of this setting again.

PR #415, opened 2019-07-01
refactor(webpack): switch to core-js from @babel/polyfill

dmbch:


dmbch:
We do not need `regenerator-runtime` as generators are [supported](https://node.green/#ES2015-functions-generators) in all node versions `untool` supports.

PR #427, opened 2019-07-11
fix: bump lodash version globally

dmbch:


dmbch:
This is needed because of `commitlint` pins the `lodash` version to 4.17.11. Will be fixed when this PR is merged..: https://github.com/conventional-changelog/commitlint/pull/715

jhiode:
Hey cool, even though it's not necessary 🙂
This only affects the local `yarn.lock` file and as we don't deploy this repo, there is no server running in production that could be vulnerable. And this `yarn.lock` file is also not part of the published packages.

dmbch:
@jhiode you are absolutely right, of course! Nevertheless, we also thought this is the right thing to do.

PR #434, opened 2019-07-24
chore: tweak release scripts and lerna config

ZauberNerd:
This commit tweaks the lerna config a bit and changes the release
scripts in our package.json.

The package.json release scripts now check for existence of `GH_TOKEN`
environment variable before any kind of release is triggered.

In normal operation a user just has to execute `yarn release` and lerna
will automatically determine the next suitable version based on the
current version _and_ also based on the current version modifier.
That means, that, if we are currently in a prerelease stage, lerna will
make sure that the next version is a prerelease too.

So in order to start an alpha release phase we need to execute
`yarn release:alpha` once and then we can continue with `yarn release`.
This will then create a version such as `2.0.0-alpha.0` and publish it
on the `next` dist-tag to npm.

Once we want to move up to the beta phase, we execute
`yarn release:beta` once and then continue with `yarn release` as
before.

The last stage would be the release candidate which we enter by
executing `yarn release:rc` once and continue with `yarn release`.

In order to move from any prerelease stage to a stable stage, we need
to execute `yarn release:graduate` which will remove the prerelease tag
and publish the version on the `latest` dist-tag to npm.


dmbch:
This looks brilliant, indeed!

PR #435, opened 2019-07-26
 feat(webpack): make max asset size configurable for target "build"

herschel666:
Here's the internal issue this PR is based on:

---
Currently teams are confronted with an omnipresent `'Some assets exceed the recommended size limit!'`-warning, quickly learning to simply ignore it. While we want teams to maintain reasonable asset sizes, we maybe don't want to follow the current, rather paternalistic approach. And we may also want to enable the teams to figure out their own individual asset size budgets and stick to them.

So my proposal would be:

- disable those warnings completely for mode `develop`
- keep it as it is for mode `node` (namely emit a warning)
- introduce a numerical config value `maxAssetSize`; if it is set, Webpack will fail with an error in mode `build`, if a chunk size exceeds that value
- keep the `maxEntrypointSize` value as it is
---

I didn't manage to have the `build` exit with status `1`. Maybe we'll have to have a look into that still. 🤔


dmbch:
By the way, I am really wondering wether we need these warnings for node builds at all. I mean, the code in question will be required/parsed/executed, anyway - and we have more specific mechanisms in place (e.g. `duplitect`) to deal with relevant issues.

dmbch:
Oh, another thought: failing the build completely (e.g. in CI) would possibly be breaking actual projects having obesity issues. This would mean that, at least in my opinion, be should only introduce such a change with the next major release.

herschel666:
Okay, here's a new, less invasive approach:

- it turns off current warnings completely
- it emits feedback only when `performance.hints` is set to `"warning"` or
`"error"` via a `configureBuild`-hook
- it keeps the red highlighting of file sizes in the `develop`- &
`build`-outputs, that exceed the set `maxAssetSize`-value

Since we don't want redundant output from Webpack, the `performance.hints`-value has to be reset to `false` after passing it into the `Logger`-plugin. This is somewhat unfortunate, but I don't see an alternative to it.

So with the current approach, we can adjust the performance hints in downstream projects freely, while also getting rid of ubiquitous max asset size warnings, that train the users to ignore them.

packages/webpack/lib/plugins/log.js
```diff
@@ -37,8 +46,10 @@ const formatSuccess = (name, duration, assets, isRebuild) => {
 };
 
 exports.LoggerPlugin = class LoggerPlugin {
-  constructor(logger) {
-    this.logger = logger;
```
dmbch: I would personally prefer separate args instead of an options hash to keep this consistent with other APIs, but I am fine with this as it is.

packages/webpack/lib/plugins/log.js
```diff
@@ -47,22 +58,33 @@ exports.LoggerPlugin = class LoggerPlugin {
       if (this.lastHashes[name] === stats.hash) {
         return;
       }
+      const { hints, maxAssetSize } = this.performance;
       const { hash, startTime, endTime } = stats;
       const isRebuild = this.lastHashes[name];
       const duration = prettyMS(endTime - startTime);
+      const { assets } = stats.toJson(statsToJsonOptions);
+      const emitAssetSizeFeedback =
```
dmbch: Variable naming appears to be a tad inconsistent here: `isRebuild` vs. `emitAssetSizeFeedback`... Should it not be `shouldEmit...`?

PR #438, opened 2019-07-30
extract common configs into shareable presets

ZauberNerd:


ZauberNerd:
This PR is marked RFC because the shareable config presets have not been published yet.

PR #442, opened 2019-08-13
Update core-js to version 3

ZauberNerd:
Even though we have this on the list for Untool 3, I think it would be useful to include it already in Untool 2.


ZauberNerd:
> I am wondering though about that untool v2/3 remark that you made - is the list you are referring to public? It might be interesting for others outside of XING as well, not just me, I guess.

Whoops, that should be untool v1/2. No, the roadmap is not yet public. We'll clean it up and put it into issues.

> Also, and possibly related: do we anticipate any sort of downstream breakage if we merge this? On the other hand, will we possibly be bitten (by bugs or missing features) if we postpone this upgrade to the next major release?

We have been bitten by missing features in core-js 2 already. We've used the `URL` global which wasn't being polyfilled.
The only possible breakage I could see would be if users were to import their own core-js functions which then would also be aliased to core-js 3.

ZauberNerd:
@dmbch @herschel666 moved this to the next branch, because we can't foresee if it would be breaking for downstream applications.

PR #453, opened 2019-08-27
feat(core): add default value syntax for env variables

herschel666:
This is an addition to the environment variable placeholder syntax, that
allows the user to set a default value. Assuming a configuration like
this:

```json
{ "value": "[ENV_VAR=fallback]" }
```

This will lead to the `"value"` being `'fallback'`, if the variable
`ENV_VAR` is not defined in the current environment.

**FYI** @dmbch @zaubernerd: I'd like to add unit tests for that. I also wonder it makes sense to limit the range of allowed characters for the default value. ~~Atm it can be literally anything while it's not possible to put a `]` in there… 🤷‍♂️~~


packages/core/lib/utils.js
```diff
@@ -75,7 +75,7 @@ exports.placeholdify = (config) => {
 exports.environmentalize = (_config) => {
   const _env = {};
   const env = global._env || process.env;
-  const regExp = /\[(\w+?)\]/g;
+  const regExp = /^\[([^=]+)=?([^\]]+)?\]$/g;
```
ZauberNerd: With the start/end anchors we restrict where an env var can be found.
Previously patterns such as this were possible: `"apiUrl": "[SERVER_URL]/api"`.
I would propose to change the regex to something like this: `\[([a-zA-Z_][a-zA-Z0-9_]*)(?:=(.+))?\]`.
I did some quick tests and it seems to work okay-ish - if that is not enough for us, we need to implement a small parse function that holds state and reads each character one by one in a loop...

ZauberNerd: Check out: https://regexr.com/4jvpr

packages/core/lib/utils.js
```diff
@@ -75,7 +75,7 @@ exports.placeholdify = (config) => {
 exports.environmentalize = (_config) => {
   const _env = {};
   const env = global._env || process.env;
-  const regExp = /\[(\w+?)\]/g;
+  const regExp = /\[([a-zA-Z_][a-zA-Z0-9_]*)(?:=(.+))?\]/g;
```
jhiode: This should be `=(.*?)` otherwise `[TEST=]` would not match the regex, which is wrong in my opinion.
And we should go for the `lazy` mode (instead of `greedy`) otherwise `[FOO=foo][BAR=bar]` would also not work.

See https://regex101.com/r/BuXyeU/1 vs https://regex101.com/r/WZhMgw/1

herschel666: Good catch! We definitely have to add unit tests here before the next major release. 😬

Issue #446, opened 2019-08-19
[RFC]: Secrets in environment variable placeholders


ZauberNerd:
We've discovered that there is a need to specify which config values should be sent to the client / or which values contain secrets and should therefore only be exposed to the server-side implementation.

There are a few solutions for white-listing values for the front-end that come to mind:
1. indicate through a naming scheme of the key, that this value should be secret (eg. `_foo`)
2. indicate through special characters in the value (eg. double brackets `[[FOO]]`) that this value should be secret
3. use an array with white-listed keys, that will be merged automatically (to allow each preset to add their own values) to indicate which values should be forwarded to the client
4. use an object that holds a mapping of keys to boolean values that should be white-listed (in contrast to the previous suggestion, this would allow users to override values set by presets)
5. allow an object as value that contains two fields (`value` and `expose` or similar) to configure that specific field

```json
{
  "name": "untest",
  "license": "MIT",
  "untool": {
    "_variant_1": "[SECRET]",
    "variant_2": "[[SECRET]]",
    "variant_3": "[NO_SECRET]",
    "whitelist_array": [
      "variant_3"
    ],
    "variant_4": "[NO_SECRET]",
    "whitelist_object": {
      "variant_4": true
    },
    "variant_5": {
      "value": "[NO_SECRET]",
      "expose": true
    }
  }
}
```

dmbch:
Another option would be to introduce an object (e.g. `serverOnly`) that would be merged and generally handled separately and that could hold the appropriate props and hidden from the client.

That being said, my personal preference would be option 1 or 2..: these appear to be most aligned with framework and ecosystem standards and conventions.

ZauberNerd:
I would prefer to white-list keys that should go to the client-side instead of black-listing them (which would be kind of the case with `serverOnly`).

I don't like variant 2 and 5 because then a user could overwrite the default value that comes from a preset and would lose the visibility settings.

With variant 1 we would need to change our code whenever the visibility changes (or build some complicated getter/proxy stuff around it), therefore I would prefer variant 3 or 4.
I think I'm leaning more towards variant 4, because that would allow later presets / the application config to overwrite the visibility of values, but since it will always be merged it is safe by default.

herschel666:
I mostly agree with @ZauberNerd here. I would add though that version 4 makes it too easy to overwrite the setting and thus provides kind of a foot gun. I also right now can't think of a situation where a preset would want to change the visibility set by a previous preset.

So I'd be in favour of variant 3.

[...]

PR #454, opened 2019-08-28
feat: only include white-listed keys in webpack bundled config

ZauberNerd:
Closes: #446


packages/webpack/preset.js
```diff
@@ -30,6 +30,13 @@ module.exports = {
       mainFields: ['mixin:server', 'mixin:runtime', 'mixin'],
     },
   },
+  browserWhitelist: {
+    host: true,
```
ZauberNerd: What about nested keys?
For example, if the config looks like this:
```json
{
  "untool": {
    "basePath": "/",
    "something": {
      "someSecret": "mysecret",
      "someSafeValue": "this-is-needed-in-the-browser"
    },
    "browserWhitelist": {
      "basePath": true
    }
  }
}
```

We could either replicate the nested object in the white-list or we could use something like the dot-notation and have the nesting entirely in the key (eg. `"something.someSafeValue": true`).

packages/webpack/mixins/config/mixin.core.js
```diff
@@ -56,6 +56,18 @@ class WebpackConfigMixin extends Mixin {
     }
     if (target === 'develop' || target === 'build') {
       configLoaderConfig.options.type = 'browser';
+      configLoaderConfig.options.config = {
+        ...this.config,
+        _config: Object.entries(this.config.browserWhitelist).reduce(
```
ZauberNerd: This only takes care of the config that is being bundled by webpack.
Another thing to watch out for would be the `_env` variables, which will still be visible in the generated HTML.
If we want to remove the corresponding `_env` entries too, we are probably going to mix client/server concerns or mix the implementation between `@untool/webpack` and `@untool/react`..

packages/webpack/preset.js
ZauberNerd: We decided to go with dot notation.

packages/react/lib/template.js
```diff
@@ -25,6 +25,7 @@ module.exports = ({
     ${fragments.noscript || ''}
     <script>
       ${Object.entries(globals)
+        .filter(([key]) => key !== '_env')
```
ZauberNerd: This takes away our ability to have environment variables at runtime (https://github.com/xing/hops#environment-variables)

packages/core/lib/config.js
```diff
@@ -25,13 +25,20 @@ exports.getConfig = ({ untoolNamespace = 'untool', ...overrides } = {}) => {
     name: pkgData.name || basename(rootDir),
     version: pkgData.version || '0.0.0',
     mixins: [],
+    browserWhitelist: {
+      rootDir: true,
```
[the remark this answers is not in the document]
ZauberNerd: `rootDir` is being used in the config loader to replace absolute paths with `.`: https://github.com/untool/untool/blob/master/packages/webpack/lib/utils/loader.js#L19
And I put the white-list here, because `rootDir` is "owned" / introduced (?) by the core (although webpack makes use of it, so maybe it's okay to move it to webpack then?).

ZauberNerd:
Only the canarist tests are failing now. I would propose to merge anyways and fix it in Hops after a new release of untool has been made.
Please take another look @jhiode @herschel666

PR #460, opened 2019-09-09
chore: set-up GH action to automatically rebase next branch

ZauberNerd:
This action will run on every push to the master branch and rebase the
next branch against master.
If the rebase fails it will check if it is fixable (ie. only conflicts
in "yarn.lock" and/or "tests/snapshots/" and then execute
"yarn install" and "yarn test -u" to fix those conflicts and continue
to rebase.

### DIFF
--- a/.github/workflows/rebase.yml
+++ b/.github/workflows/rebase.yml
@@ -1,0 +1,20 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  rebase:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Rebase next against master
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          git remote set-url origin https://x-access-token:$GITHUB_TOKEN@github.com/untool/untool.git
+          git config --global user.email "action@github.com"
+          git config --global user.name "GitHub Action"
+          ./scripts/rebase.sh master next
--- a/scripts/rebase.sh
+++ b/scripts/rebase.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+
+set -ex
+
+base_brach=${1:-master}
+head_branch=${2:-next}
+
+git fetch origin "$base_brach"
+git fetch origin "$head_branch"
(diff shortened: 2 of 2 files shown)


.github/workflows/rebase.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  rebase:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Rebase next against master
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          git remote set-url origin https://x-access-token:$GITHUB_TOKEN@github.com/xing/hops.git
```
herschel666: shouldn't this be `…untool/untool.git`?!

scripts/rebase.sh
```diff
@@ -0,0 +1,38 @@
...
+base_brach=${1:-master}
+head_branch=${2:-next}
+
+git fetch origin "$base_brach"
+git fetch origin "$head_branch"
+git checkout -b "$head_branch" "origin/$head_branch"
+
+is_rebasing=1
+
+exitcode=0 && git rebase "origin/$base_brach" || exitcode=$?
```
herschel666: why chaining the setting of `exitcode` to the rebasing & why setting it at all, and not just using `$?`?

scripts/rebase.sh
```diff
@@ -0,0 +1,38 @@
...
+    yarn install
+    yarn test -u
+    git add yarn.lock tests/snapshots
+    exitcode=0 && git rebase --continue || exitcode=$?
+    if [[ $exitcode -eq 0 ]]; then
+      is_rebasing=0
+      break
+    fi
+  else
+    break
```
herschel666: why bother to break at the very end of the loop?

scripts/rebase.sh
ZauberNerd: Because I want to keep the exit code and if I were just using `$?` it would be the exit code of the last command, which might not be what I'm interested in.

scripts/rebase.sh
ZauberNerd: Because `$is_rebasing` might still be set to `1` but then it is a conflict that can't be resolved automatically, therefore I have to break out of the loop.

PR #465, opened 2019-10-01
feat(express): add cookie-parser middleware

jhiode:
This adds the `cookie-parser` middleware to express.
I think we already talked (and agreed) about this; it makes the life in Hops-land a little bit easier.

And maybe we should back-port the fix `helmet` for 1.x as well.


packages/express/mixins/mixin.core.js
```diff
@@ -42,8 +42,10 @@ class ExpressMixin extends Mixin {
       const helmet = require('helmet');
       const express = require('express');
       const mime = require('mime');
+      const cookieParser = require('cookie-parser');
       const { distDir } = this.config;
-      middlewares.initial.push(helmet());
+      middlewares.preinitial.push(helmet());
```
herschel666: I wonder whether I'd prefer…??

```js
middlewares.preinitial.push(helmet(), cookieParser());
```

jhiode:
Do we consider the middleware phase change as breaking?

PR #472, opened 2019-10-15
feat: replace react-helmet with react-helmet-async

ZauberNerd:
The https://github.com/nfl/react-helmet repository seems fairly dead,
there are a lot of open issues and PRs and not many releases / commits
recently.
Also the author of the underlying `react-side-effect` library
discourages its usage: https://github.com/gaearon/react-side-effect/issues/54#issuecomment-520056590

The npm numbers verify, that react-helmet-async now has more downloads
than react-helmet:
- https://www.npmjs.com/package/react-helmet
- https://www.npmjs.com/package/react-helmet-async

Read more:
- https://github.com/nfl/react-helmet/issues/465
- https://github.com/gaearon/react-side-effect/issues/40


packages/react/lib/fragments.js
```diff
@@ -1,13 +1,6 @@
...
-} = require('react-helmet');
 
 module.exports = (element) => {
   const reactMarkup = renderToString(element);
-  const fragments = Object.entries(renderStatic()).reduce(
-    (result, [key, value]) => ({ ...result, [key]: value.toString() }),
-    { reactMarkup, headPrefix: '', headSuffix: '' }
-  );
-  return fragments;
+  return { reactMarkup };
```
ZauberNerd: Should we delete this function / file?
We could potentially also delete the `renderToFragments` hook, because it is no longer needed (only hops-graphql relies on it and it also is not documented).

packages/react/lib/react-helmet-shim.js
```diff
@@ -0,0 +1,15 @@
+const deprecate = require('depd')('@untool/react');
+const reactHelmetAsync = require('react-helmet-async');
+
+deprecate(
+  '[DEP004] react-helmet is deprecated. Please switch to react-helmet-async instead (https://github.com/untool/untool/blob/master/DEPRECATIONS.md#dep004).'
+);
+
+module.exports = reactHelmetAsync;
+module.exports.default = reactHelmetAsync.Helmet;
```
ZauberNerd: This will probably break for apps that are written in commonjs. Is that a use-case that we want to support as well?

jhiode: react-helmet (v5.2.1) is doing this:
```js
exports.__esModule = true;
...
exports.Helmet = HelmetExport;
exports.default = HelmetExport;
```
(https://unpkg.com/react-helmet@5.2.1/lib/Helmet.js)

And
```js
export { HelmetExport as Helmet };
export default HelmetExport;
```
(https://unpkg.com/react-helmet@5.2.1/es/Helmet.js)

So I think we should do the same here:
```js
exports.__esModule = true;
exports.Helmet = reactHelmetAsync.Helmet;
exports.default = reactHelmetAsync.Helmet;
```

packages/react/lib/fragments.js
herschel666: We can definitely delete this function, but aren't we relying on the hook in [Apollo SSR](https://github.com/xing/hops/blob/6e5adbcdcbb6c54a9c2bb03dff1965cb9d0a3e7f/packages/graphql/mixin.server.js#L98)?

packages/react/lib/react-helmet-shim.js
```diff
@@ -0,0 +1,16 @@
...
+);
+
+module.exports = exports = reactHelmetAsync;
+exports.__esModule = true;
+exports.default = reactHelmetAsync.Helmet;
+
+deprecate.property(
+  module.exports,
+  'default',
+  '[DEP005] The default export of react-helmet-async is deprecated. Please use the named export instead (https://github.com/untool/untool/blob/master/DEPRECATIONS.md#dep005).'
```
jhiode: Hm, I don't understand this depreciation warning.
Isn't this the export for:`import Helmet from 'react-helmet'`? So why isn't this just the same warning as `DEP004`?

ZauberNerd: `react-helmet-async` originally does not provide a default export, while `react-helmet` does.
And we still have a lot of code internally that does `import Helmet from 'react-helmet';` which would then fail. Therefore we provide the default alias ourselves, but at the same time print a deprecation warning so that users switch to using the named export `import { Helmet } from 'react-helmet-async';`.

ZauberNerd: Updated.

packages/react/lib/fragments.js
ZauberNerd: It is also being imported in Hops: https://github.com/xing/hops/blob/6e5adbcdcbb6c54a9c2bb03dff1965cb9d0a3e7f/packages/graphql/mixin.server.js#L2 Let's leave it in for now, and discuss if/how we want to change the mixins in a different PR.

PR #479, opened 2019-11-06
fix: correct build script to be platform neutral

KnisterPeter:


package.json
```diff
@@ -23,7 +23,7 @@
     "reset": "git clean -dfx && yarn",
     "pretest": "git clean -dfX tests/fixtures/untest",
     "posttest": "git clean -dfXq tests/fixtures",
-    "preinstall": "test -z \"${npm_execpath##*yarn*}\""
+    "preinstall": "node -e \"process.exit(process.env.npm_execpath.indexOf('yarn') > -1 ? 0 : 1)\""
```
ZauberNerd:
```suggestion
    "preinstall": "node -e \"process.exit(process.env.npm_execpath.includes('yarn'))\""
```
would work as well :)

KnisterPeter: I opted for well defined exit codes `0` and `1` instead of `true` or `false` 😉 
And I'm not sure if `includes` is supported on all node versions this might be installed with (e.g. 8 or 10?)

ZauberNerd: It is supported (https://node.green/#ES2015-built-in-extensions-String-prototype-methods), but yeah, this is fine too, if you want to merge it as is.

herschel666: `.includes(…)` is supported in Node 8 afaict. And I'd propose to go with `.include` **&** the ternary operator.

PR #501, opened 2019-12-04
chore: point canarist to branch "v11.x" of Hops

herschel666:


herschel666:
Closing because we want to go _release-branch_ here, too.

PR #504, opened 2019-12-05
update babel packages to v7.7

jhiode:
This is a preparation for https://github.com/xing/hops/pull/969


herschel666:
In which release branches is this supposed to end up?

PR #528, opened 2020-01-08
chore: allow renovate to update release branches

KnisterPeter:


package.json
```diff
@@ -109,6 +109,24 @@
   "renovate": {
     "extends": [
       "@untool/renovate-config:base"
+    ],
+    "baseBranches": [
+      "master",
+      "v12.x",
```
ZauberNerd: These should be v1 and v2

KnisterPeter: 👍 That was just a test 🤣

PR #542, opened 2020-01-14
fix(core): align mixin load order with config overrides

jhiode:
At the moment we have a logic to merge mixins together but that order is actually different as for config values. So if we use mixin hooks with the `override` strategy the override does not happen in the same order as config values.

Example:
We have three packages: a, b, and c.

```js
// a/preset.js:
{
  presets: ['b'],
  test: 'a',
}
// a/mixin.core.js:
class AMixin extends Mixin {
  hook() {
    return 'a';
  }
}
```
```js
// b/preset.js:
{
  test: 'b',
}
// b/mixin.core.js:
class BMixin extends Mixin {
  hook() {
    return 'b';
  }
}
```
```js
// c/preset.js:
{
  presets: ['a', 'b'],
}
// c/mixin.core.js:
class CMixin extends Mixin {
  test() {
    console.log(this.hook());
  }
}
```

Calling `hook` would return `'a'` while the config value for `test` is `'b'`.
The reason for that is that the mixins stay in the same location in the array even if they are required another time while config values are just getting overridden again.

Loading order for this example is:
```
/path/to/b/
/path/to/a/
/path/to/c/
```

This PR changes the order to:
```
/path/to/a/
/path/to/b/
/path/to/c/
```

I understand that there might be reasons why the current behaviour would be preferable but I think it's too confusing if the mixins have their own rules and differ from the config behaviour.


ZauberNerd:
Added a config option `enableLegacyMixinSortOrder` to revert back to the old mixin config order.
This defaults to `false`.

PR #560, opened 2020-01-27
Handle functions as config values appropriately

herschel666:


herschel666:
btw: we should imho put this rule ([`7e25605`](https://github.com/untool/untool/pull/560/commits/7e25605025e51a9e929b3b8898cf2dbbec9c6f5f)) into our eslint config.

packages/core/lib/utils.js
```diff
@@ -80,7 +80,7 @@ exports.placeholdify = (config) => {
         {}
       );
     }
-    if (regExp.test(item)) {
+    if (typeof item === 'string' && regExp.test(item)) {
```
KnisterPeter: How about adding a test case for this?

herschel666: Good idea!

herschel666: Done.

